PR #6, opened 2016-10-28
PAE-41 - Panel init callback is called before the panel is initialized

francescorivola:


zyhfish:
changes loooks good and merged.

PR #8, opened 2016-11-11
install module in different folder.

zyhfish:


ashishpd:
This has been implemented long time back via diff set of commits. Closing.

PR #20, opened 2017-01-10
CONTENT-7833 The "Cancel" buttons on the Install always cancel out of…

pberna:
Could you please review my changes?

### DIFF
--- a/src/Modules/Settings/Dnn.PersonaBar.Extensions/Extensions.Web/src/components/InstallExtensionModal/index.jsx
+++ b/src/Modules/Settings/Dnn.PersonaBar.Extensions/Extensions.Web/src/components/InstallExtensionModal/index.jsx
@@ -111,23 +111,12 @@ class InstallExtensionModal extends Component {
         });
     }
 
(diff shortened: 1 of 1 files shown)


src/Modules/Settings/Dnn.PersonaBar.Extensions/Extensions.Web/src/components/InstallExtensionModal/index.jsx
```diff
@@ -128,6 +128,7 @@ class InstallExtensionModal extends Component {
             selectedLegacyType: null,
             alreadyInstalled: false
         });
+        props.onCancel();
```
francescorivola, line 120: @mgatmaytan, can you please help me on this review. The cancelInstall receive a cancelRevertStep but it look like this is always set to false + for some reason the callback in clearParsedInstallationPackage is not called or does not work. Can you please help on review the @pberna fix and cleanup as part of this PR if those code is not needed. Thks.

francescorivola (changes requested):

PR #25, opened 2017-01-29
Removed debug URL in pages extension

janjonas:
There is no need for this JS comment in a production environment.


ashishpd:
For debugging using webpack, we actually use this URL. Having it commented is v handy. Sorry, I have to reject this.

janjonas:
@ashishpd: Thanks for the quick reply. I assumed that this is the webpack URL that is needed during development. But I disagree: This is nothing that should be checked nor should this line be present in a production environment. There are 14 other similar code fragments in the Dnn.AdminExperience.Extensions
repository and only this one contains the debug URL. In my project I configure the URL in the backend and pass it as option to the PeronaBar frontend. IMHO, this is much cleaner and not so error-prone (checkin in the webpack URL breaks the module).

PR #35, opened 2017-03-08
DNN-9526: ADA Compliance issues - PB Users.

zyhfish:


src/Modules/Manage/Dnn.PersonaBar.Users/Users.Web/src/_exportables/src/components/UserTable/UsersRoles/index.jsx
```diff
@@ -191,6 +198,7 @@ class UserRoles extends Component {
                     <GridCell columnSize={50}>
                         <span>
                             <Combobox suggest={false}
+                                ref={(dom) => {this.comboBoxDom = dom;}}
```
kanm2015, line 201: usually we don't wanna make DOM operations out of React dataflow but I think this is fine for now. Later we can use our own common component rather than React Combobox

kanm2015:
Code review done

PR #45, opened 2017-03-16
DNN-9363: remove tab urls when drag new add page.

zyhfish:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/IPagesController.cs
```diff
@@ -18,6 +18,8 @@ public interface IPagesController
 
         void DeletePage(PageItem page);
 
+        void InitializePage(int pageId);
```
francescorivola: Add some metadata doc to this method explaining the goal of the method

src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageHierarchy/pages.pageHierarchy.js
```diff
@@ -725,6 +725,7 @@ window.dnn.pages = window.dnn.pages || {};
                         sourceFind   = handler._findDataPosition(sourcePageId);
                         source       = viewModel.pagesList()[sourceFind.level].pages()[sourceFind.index];
                         sourceData   = handler._clonePageData(source);
+                        isDragItem = item.find('> div').hasClass('drag-item');
```
francescorivola: is this only true if is new page? We may want put some different name so is clear the scenario where this got true and we perform the initialize request.

src/Modules/Content/Dnn.PersonaBar.Pages/Services/Dto/PageMoveRequest.cs
```diff
@@ -9,5 +9,7 @@ public class PageMoveRequest
         public int ParentId { get; set; }
 
         public string Action { get; set; }
+
+        public bool Initialize { get; set; }
```
francescorivola: add metadata doc

francescorivola (changes requested):

PR #46, opened 2017-03-16
DNN-9540: Fix portal alias cannot contain dash issue

kanm2015:


zyhfish:
changes looks good and merged.

PR #71, opened 2017-04-18
DNN-9190: Keep script tag on page header tags

kanm2015:


zyhfish:
changes looks good and merged.

PR #76, opened 2017-04-20
DNN-9700: Fixed alignment issue for long text

kanm2015:


zyhfish:

PR #77, opened 2017-04-20
Remove duplicate keys in SiteImportExport.resx

valadas:
The fooloing keys where defined twice in the file:

ImportSummary.Text
Assets.Text
Users.Text
Roles.Text
Vocabularies.Text
CancelImport.Text


kanm2015:
change looks good, PR merged, thx

PR #94, opened 2017-04-26
DNN-9274: fixing listing deleted superusers in PB.

galatrash:


amarjit-dhunna:
Changes not required as the deleted super user should be visible in superusers filter only

PR #96, opened 2017-04-28
DNN-9713: Updating instead of removing portal localization entry when…

kanm2015:
… disabling content localization


zyhfish:

PR #102, opened 2017-05-10
DNN-9759 Add "Number of Days Before Password Reuse" setting

bdukes:
This is the UI implementation for dnnsoftware/Dnn.Platform#1727


bdukes:
I've had no luck getting this to compile, so I can't guarantee the completeness of this PR, but I think it should be most of the way there.

On the C# side, seeing errors like this:
>Unable to find version '9.1.1.33' of package 'DotNetNuke.Bundle'

On the JS side, seeing errors like this:
> An unexpected error occurred: "https://www.myget.org/F/dnn-software-private/npm/array.prototype.find/-/array.prototype.find-2.0.4.tgz: Request failed \"401 Unauthorized\"".

valadas:
I am also not able to compile the admin experience project with the same error, is there any help for the community, a tutorial maybe would be nice so we can publish correct pull requests. I have logged quite a few bugs that I could provide the fix but because of the fact I can't compile the project, I am blocked. Or is there an issue with Myget maybe?

zyhfish:
Hi @bdukes @valadas , thanks for your contribution! you can reference to http://www.dnnsoftware.com/community-blog/cid/155391/dnn9-adding-a-reactjs-app-to-persona-bar to config the envrionment and let bundle js can be generated successful.
for the C# project complilation, you can unload the NugetPackages project, then copy needed assemblies into src\bin folder, then the project should able to build successful. also we have source packages for this repo during each build, which have the needed assemblies in src\bin folder already. if you need that package, please let us know, we can upload them into fixed path so that you all can get it and try. thx

zyhfish (changes requested):
@bdukes , can you please resolve the conflicts and do the test in your local side base on my previous comment, if you need more help, just ping me and i will reply ASAP. thx

bdukes:
@zyhfish I was able to get this to build, and I rebased to resolve conflicts.  Let me know if there's more you need from me

kanm2015:
code review done

PR #103, opened 2017-05-11
DNN-9762: add edit permission.

zyhfish:


kanm2015:
dev tests done, works as expected, PR merged

PR #114, opened 2017-05-30
DNN-9125: use friendly text to display remaining time

kanm2015:


zyhfish:

PR #124, opened 2017-06-06
DNN-9895: Updated to return localized msg for save/delete api and rem…

kanm2015:
…oved dup entries in resx


zyhfish:

PR #128, opened 2017-06-07
DNN-9755: Fixed switches not localized issue

kanm2015:


zyhfish:

PR #132, opened 2017-06-09
DNN-9905: added validation on more fields

kanm2015:


zyhfish:

PR #134, opened 2017-06-12
DNN-9688: Added missing functionality for managing entries of list ty…

kanm2015:
…pe profile property


zyhfish:

PR #138, opened 2017-06-15
Contribute Prompt Command Line to DNN Core

dnndev:
This is the Prompt command line project to add to DNN Core. 
Reference: https://dnntracker.atlassian.net/browse/DNN-9917

Prompt as it stands now, offers CRUD-type operations for many DNN components like Users, Pages, Roles, etc.  It uses a consistent `verb-object` syntax familiar to Powershell. That syntax is documented in src/Modules/Prompt/Dnn.PersonaBar.Prompt/admin/personaBar/help/learn.html. Not all commands and not all options for commands are (or necessarily should be) implemented.

Key design goals are:

1) Minimize typing without sacrificing understanding: Keep command names short and consistent without using obscure abbreviations or flags. Leverage the HTML environment when it benefits the user (as in hyperlinks in command results that generate commands), while remaining true to the speed and simplicity of the command line.

2) Minimize the learning curve and intimidation of command lines as much as possible through the use of in-depth, easily accessible help that details syntax, usage notes, and provide examples.

3) Make Prompt easily extensible so that developers can provide their own commands. This is accomplished by Prompt handling the UI so that the extension only needs to provide the functionality and resulting data.

Notes about the code:

1) The client-side functionality part of prompt can be found in /src/Modules/Prompt/Prompt.Web/ 
2) We use webpack to bundle and package.
3) Help files are found here: /src/Modules/Prompt/Dnn.PersonaBar.Prompt/admin/personaBar/help

Obviously, you will need to make changes and incorporate the project into your workflow, but we tried to mimic the existing Persona Bar project structure as much as possible.

We're happy to answer any questions and work with you to make the integration as smooth as possible.


ashishpd:
@amarjit-dhunna  Please merge this as is, but don't add the project to the solution. We don't want it to be built on our build servers just yet. We will refactor the code after you merge. The implementation of the Interface will be moved to specific Persona Bar extensions. Currently everything is implemented at one place. We will also move the interface to Admin Library (to help with dependencies).  Prompt will be an PB extension as well. One can choose to not install or uninstall after the fact.

PR #139, opened 2017-06-15
DNN-9535: Fixed double html encoding issue when saving resx value

kanm2015:


zyhfish:

PR #142, opened 2017-06-19
CONTENT-7660: Use issecure parameter while create tab from template

amarjit-dhunna:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/TemplateController.cs
```diff
@@ -156,7 +156,7 @@ public void CreatePageFromTemplate(int templateId, TabInfo tab, int portalId)
                 var xmlDoc = new XmlDocument();
                 try
                 {
-                    // open the XML file
+                    // open the XML filew
```
kanm2015: >_< filew

kanm2015:
code review done

PR #151, opened 2017-06-23
DNN-9919: show error msg for both add & save methods

kanm2015:


zyhfish:

PR #153, opened 2017-06-23
Bug/dnn 9953

kanm2015:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/PageUrlsController.cs
```diff
@@ -149,15 +149,18 @@ public PageUrlResult UpdateCustomUrl(SaveUrlDto dto, TabInfo tab)
             }
 
             //Validate for uniqueness
-            urlPath = FriendlyUrlController.ValidateUrl(urlPath, -1, portalSettings, out modified);
```
amarjit-dhunna, line 152: @kanm2015 , It should be enough to just use tab.TabID here instead of -1 in the FriendlyUrlController.ValidateUrl method. We don't need any more code changes I beleive.

kanm2015:
will fix the issue in other PR, close this one for now

PR #154, opened 2017-06-23
DNN-9821 - Increases size of header title when in textoverflow mode

danieldram:


kanm2015:
close this PR since we will leave the size as is and remove "Page Settings:" from header

PR #163, opened 2017-06-30
DNN-9982: fixed passing wrong portalId to page picker issue

kanm2015:


amarjit-dhunna:
Changes good and merged

PR #169, opened 2017-07-04
DNN-9953: updated to allow editing urls and prevent from adding dupli…

kanm2015:
…cate urls


zyhfish:

PR #177, opened 2017-07-14
DNN-10018: Updated GetPageList to support more filters

kanm2015:


kanm2015:
will combine it w/ some other changes, close for now

PR #187, opened 2017-07-21
DNN-10064: added rsvp code support on roles

kanm2015:


zyhfish:

PR #198, opened 2017-07-28
DNN-8941: packages names checks should be case insensitive.

galatrash:
Changes to use "OrdinalIgnoreCase" instead of "InvariantCultureIgnoreCase" was based on this:
https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings


kanm2015:
code review done

PR #211, opened 2017-08-03
DNN-9963: added svg to the supported image file types

kanm2015:


zyhfish:

PR #229, opened 2017-09-07
DNN-10166 IP Filters is broken

tpluscode:
Modified the back-end code to return Bad Request when IP address or submask are not valid.
Improved regular expression for front-end validation

### DIFF
--- a/src/Modules/Settings/Dnn.PersonaBar.Security/Security.Web/package.json
+++ b/src/Modules/Settings/Dnn.PersonaBar.Security/Security.Web/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "set NODE_ENV=production&&webpack -p",
-    "webpack": "webpack-dev-server -d --port 8080 --hot --inline --content-base dist/ --history-api-fallback",
+    "webpack": "webpack-dev-server -d --host 0.0.0.0 --port 8080 --hot --inline --content-base dist/ --history-api-fallback",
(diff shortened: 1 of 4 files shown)


src/Modules/Settings/Dnn.PersonaBar.Security/Security.Web/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "set NODE_ENV=production&&webpack -p",
-    "webpack": "webpack-dev-server -d --port 8080 --hot --inline --content-base dist/ --history-api-fallback",
+    "webpack": "webpack-dev-server -d --host 0.0.0.0 --port 8080 --hot --inline --content-base dist/ --history-api-fallback",
```
tpluscode, line 7: I added this parameter here so that the dev-server can be accessed from virtual machine (running VMWare on Mac). [This answer](https://apple.stackexchange.com/a/180636/253832) explains why it's necessary.

src/Modules/Settings/Dnn.PersonaBar.Security/Security.Web/src/components/auditCheck/index.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-danger */
```
tpluscode, line 1: This is an unrelated change, but we needed that to fix that to have webpack compile successfully.

zyhfish:

PR #230, opened 2017-09-07
DNN-10154 - fixed super users search issue

vano-maisuradze:
Added @IncludeSuperUsers parameter in Personabar_GetUsers stored procedure to include/exclude super users from search result.
Also Lucene search was always excluding super users from search. Now superuser keyword is added in NumericKeys only if user is no super user.


sleupold:
Please do not use WITH (NoLock) in your SQL queries!

fjlucas:
Thank @sleupold  for your comment, but in order to not introduce possible side effects in the issue, we've just focus on fixing the bug. The use of WITH comes from the previous implementation and it's out of the scope of this issue.

Finally the code looks good.

PR #231, opened 2017-09-07
DNN-10163 - AddLogSettings now returns 401 status if user has admin role

vano-maisuradze:
AddLogSettings will return 401 http status code, if user has admin role.


fjlucas:
Code looks good.

PR #235, opened 2017-09-11
DNN-8460: Updated to return localized error msg for creating invalid …

kanm2015:
…page name


zyhfish:

PR #241, opened 2017-09-18
DNN-9638: Updated API to copy selected modules only

kanm2015:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -1224,67 +1224,66 @@ private void CopySourceTabProperties(TabInfo tab, TabInfo sourceTab)
...
+                    newModule.TabID = tab.TabID;
+                    newModule.DefaultLanguageGuid = Null.NullGuid;
+                    newModule.CultureCode = tab.CultureCode;
+                    newModule.ModuleTitle = module.Title;
+
+                    if (copyType != ModuleCopyType.Reference)
+                    {
+                        newModule.ModuleID = Null.NullInteger;
+                        ModuleController.Instance.InitialModulePermission(newModule, newModule.TabID, 0);
+                    }
```
amarjit-dhunna: I think we need to have the InheritViewPermissions copied as well as the Module Permissions. Please check ManageTabs.ascx.cs:1030 from 8.0.4 release branch of Platform

PR #253, opened 2017-09-27
DNN-9890: returns null for GetInternalUrl if no file id

kanm2015:


zyhfish:

PR #289, opened 2017-10-05
CONTENT-7976: User parent page permissions when adding as child from …

amarjit-dhunna:
…In Context menu.


kanm2015:
code review done

PR #293, opened 2017-10-06
DNN-10312: updated UI to prompt error and change 500 error to 403 if …

kanm2015:
…no permission to change parent


kanm2015:
merged

PR #295, opened 2017-10-10
DNN-8875: order the alias list to make primary alias at top of list a…

aliazizoglu:
…nd then select the first one.


kanm2015:
code review done

PR #297, opened 2017-10-10
DNN-10045: Fix security issues when creating page with a page with no…

amarjit-dhunna:
… permissions


kanm2015:
code review done

PR #300, opened 2017-10-11
DNN-10272: do not redirect if page type is not normal or page is disa…

kanm2015:
…bled, update treeview and select new added page as current page

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/pageActions.js
+++ b/src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/pageActions.js
@@ -240,11 +240,12 @@ const pageActions = {
                 if (selectedPage.tabId > 0) {
                     utils.notify(Localization.get("PageUpdatedMessage"));
                 }
-                if (response.Page.canAddContentToPage)
+                if (response.Page.canAddContentToPage && selectedPage.pageType === "normal" && !selectedPage.disableLink) {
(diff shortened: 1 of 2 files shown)


src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/pageActions.js
```diff
@@ -240,11 +240,12 @@ const pageActions = {
...
                     utils.notify(Localization.get("PageUpdatedMessage"));
                 }
-                if (response.Page.canAddContentToPage)
+                if (response.Page.canAddContentToPage && selectedPage.pageType === "normal" && !selectedPage.disableLink) {
                     PagesService.openPageInEditMode(response.Page.id, response.Page.url);
+                }
                 else if (typeof callback === "function") {
                     utils.notify(Localization.get("PageCreatedMessage"));
-                    callback();
+                    callback(response.Page);
```
amarjit-dhunna, line 248: Response.Page should be added to the pages list only if Response.page.canAddContentToPage is true and Response.page.publishStatus is published. If both conditions do not fulfill, the page is not supposed to show up in the list.
Note: page.publishStatus is added in DNN-10127, so this branch will need to fetch changes from DNN-10127 or should be merged after DNN-10127 is merged.

kanm2015: don't need to check canAddContentToPage, will check publishStatus

PR #301, opened 2017-10-12
DNN-10299: SECURITY: Admin Security settings issues for API level calls

Mohtshum:
- Validation added by attribute filter so that It could be used prior to Model objects on both UpdateRegistrationSettingsRequest & UpdateBasicLoginSettingsRequest
- Localization messages are also added with format "{0}  {1} is not valid." i.e. if wrong culturecode added like pkr then it would be "Culturecode pkr is not valid."
- Nuget Package settings are also added so that it could be compilable
- Please give feedback on location of new classes added regarding folder used
- Though these validations are added for API calls but also applicable for application frontend
- Hard coded logics were also moved to separate helper class namely RegistrationSettingsHelper so that can be used for validation logic

### DIFF
--- a/src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
+++ b/src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
@@ -27,7 +27,7 @@ using DotNetNuke.Services.Localization;
 
 namespace Dnn.PersonaBar.Security.Attributes
 {
-    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Property)]
     class AuthProviderSupportedAttribute : ValidationAttribute
     {
         protected override ValidationResult IsValid(object value, ValidationContext validationContext)
@@ -36,15 +36,15 @@ namespace Dnn.PersonaBar.Security.Attributes
 
             if (string.IsNullOrWhiteSpace(value.ToString()))
             {
-                return new ValidationResult(string.Format(Localization.GetString(Constants.NotValid + ".Text", Constants.LocalResourcesFile), propertyName, value.ToString()));
+                return new ValidationResult(string.Format(Localization.GetString(Constants.EmptyValue, Constants.LocalResourcesFile), propertyName));
             }
 
             var allAuthProviders = new SecurityController().GetAuthenticationProviders();
 
             if (!allAuthProviders.Contains(value))
             {
-                return new ValidationResult(string.Format(Localization.GetString(Constants.NotValid + ".Text", Constants.LocalResourcesFile), propertyName, value.ToString()));
-            }            
+                return new ValidationResult(string.Format(Localization.GetString(Constants.NotValid, Constants.LocalResourcesFile), propertyName, value.ToString()));
+            }
 
             return ValidationResult.Success;
         }
--- a/src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
+++ b/src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
@@ -1,0 +1,52 @@
+﻿#region Copyright
+// 
+// DotNetNuke® - http://www.dotnetnuke.com
+// Copyright (c) 2002-2017
+// by DotNetNuke Corporation
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated 
+// documentation files (the "Software"), to deal in the Software without restriction, including without limitation 
+// the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and 
+// to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions 
+// of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED 
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL 
+// THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF 
+// CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
+// DEALINGS IN THE SOFTWARE.
+#endregion
+
+using System;
+using System.Linq;
+using System.ComponentModel.DataAnnotations;
+using Dnn.PersonaBar.Security.Components;
+using DotNetNuke.Services.Localization;
+
+namespace Dnn.PersonaBar.Security.Attributes
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    class AuthProviderSupportedAttribute : ValidationAttribute
+    {
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            var propertyName = validationContext.DisplayName;
+
+            if (string.IsNullOrWhiteSpace(value.ToString()))
+            {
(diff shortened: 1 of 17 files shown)


src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (string.IsNullOrWhiteSpace(value.ToString()))
+            {
+                return new ValidationResult(string.Format(Localization.GetString(Constants.NotValid + ".Text", Constants.LocalResourcesFile), propertyName, value.ToString()));
+            }
+
+            var allAuthProviders = new SecurityController().GetAuthenticationProviders();
+
+            if (!allAuthProviders.Contains(value))
+            {
+                return new ValidationResult(string.Format(Localization.GetString(Constants.NotValid + ".Text", Constants.LocalResourcesFile), propertyName, value.ToString()));
```
Mohtshum: Please check this entry for resource file for localization

src/Modules/Settings/Dnn.PersonaBar.Security/Helper/RegistrationSettingsHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+// CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
+// DEALINGS IN THE SOFTWARE.
+#endregion
+
+using DotNetNuke.Services.Localization;
+using System.Collections.Generic;
+
+namespace Dnn.PersonaBar.Security.Helper
+{
+    class RegistrationSettingsHelper
```
Mohtshum, line 27: These hard coded logic are moved From GetRegistrationSettings of Services/SecurityController class so that to re-use for validation in Attribute filter. Please suggest any better way if exists.

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
kanm2015: no need to append ".Text", Localization.GetString() is able to get that

kanm2015: can we get rid of the extra space between {0} and {1} in entry "{0}  {1} is not valid."?

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/CultureCodeExistAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    class CultureCodeExistAttribute : ValidationAttribute
+    {
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            var propertyName = validationContext.DisplayName;
+
+            if (string.IsNullOrWhiteSpace(value.ToString()))
+            {
+                return new ValidationResult(string.Format(Localization.GetString(Components.Constants.NotValid + ".Text",Components.Constants.LocalResourcesFile),propertyName, value.ToString()));
```
kanm2015: can we have a different error msg for empty value?

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/CultureCodeExistAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+
+            var allLocales = LocaleController.Instance.GetLocales(PortalController.Instance.GetCurrentPortalSettings().PortalId);
+
+            if (!allLocales.Select(l=>l.Value.Code).Contains(value.ToString()))
```
kanm2015: let's compare case insensitive

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System;
+using System.Linq;
+using System.ComponentModel.DataAnnotations;
+using Dnn.PersonaBar.Security.Components;
+using DotNetNuke.Services.Localization;
+
+namespace Dnn.PersonaBar.Security.Attributes
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
```
kanm2015: we don't need AllowMultiple attribute here since it's the default value.

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/CultureCodeExistAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using DotNetNuke.Services.Localization;
+using System;
+using System.Linq;
+using System.ComponentModel.DataAnnotations;
+using DotNetNuke.Entities.Portals;
+
+namespace Dnn.PersonaBar.Security.Attributes
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
```
kanm2015: remove Allowmultiple

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/AuthProviderSupportedAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+    class AuthProviderSupportedAttribute : ValidationAttribute
+    {
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            var propertyName = validationContext.DisplayName;
+
+            if (string.IsNullOrWhiteSpace(value.ToString()))
+            {
+                return new ValidationResult(string.Format(Localization.GetString(Constants.NotValid + ".Text", Constants.LocalResourcesFile), propertyName, value.ToString()));
```
kanm2015: can we have a different error msg for empty value?

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/RegistrationFieldsAttribute.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public RegistrationFieldsAttribute(string registrationFormType, string requireUniqueDisplayName)
+        {
+            RegistrationFormTypePropertyName = registrationFormType;
+            RequireUniqueDisplayNamePropertyName = requireUniqueDisplayName;
+        }
+
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            var registrationFormTypeValue = validationContext.ObjectType.GetProperty(RegistrationFormTypePropertyName).GetValue(validationContext.ObjectInstance, null).ToString();
```
kanm2015: let's add a null check for validationContext.ObjectType.GetProperty(RegistrationFormTypePropertyName)

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/UserExistAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    class UserExistAttribute : ValidationAttribute
+    {
+        public string[] RoleNames { get; set; }
+
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            var propertyName = validationContext.DisplayName;
+            int UserId;
```
kanm2015: userId? let's follow the naming convention of lower camel case

src/Modules/Settings/Dnn.PersonaBar.Security/Helper/RegistrationSettingsHelper.cs
kanm2015: looks good

src/Modules/Settings/Dnn.PersonaBar.Security/Attributes/TabExistAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            var propertyName = validationContext.DisplayName;
+            int tabId = Convert.ToInt32(value.ToString());
+
+            if ( tabId != -1)
+            {
+                var portalSetting = PortalController.Instance.GetCurrentPortalSettings();
+                var tab = TabController.Instance.GetTab(tabId, portalSetting.PortalId);
+
+                if (tab == null)
```
kanm2015, line 43: let's also make sure the tab is not disabled or soft deleted

kanm2015:
code review done

PR #330, opened 2017-10-20
DNN-10379: show confirmation before switching to a no permission page

kanm2015:


src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/App.jsx
```diff
@@ -725,12 +725,20 @@ class App extends Component {
...
     onNoPermissionSelection(pageId) {
-        this.noPermissionSelectionPageId = pageId;
+        const { selectedPage, selectedPageDirty } = this.props;
+        this.selectPageSettingTab(0);
+        this.shouldRunRecursive = false;
+        const left = () => { 
+            this.noPermissionSelectionPageId = pageId;
+            this.setEmptyStateMessage(Localization.get("NoPermissionEditPage"));
+        };
+        const right = () => (!selectedPage || pageId !== selectedPage.tabId) ? this.showCancelWithoutSavingDialogInEditMode(pageId) : null;
```
amarjit-dhunna: @kanm2015 Doesn't matter if page has changes or not, the page selection should still be based upon permissions. The onNoPermissionSelection method is supposed to show "NoPermission" message after the Confirmation box and should not make call to GetPageDetails as that will fail. You must set this.noPermissionSelectionPageId = pageId in both cases and nothing else should happen after confirmation is said as continue.

PR #331, opened 2017-10-23
CONTENT-8480: use master portal id to get users.

zyhfish:


src/Modules/Manage/Dnn.PersonaBar.Users/Services/UsersController.cs
```diff
@@ -114,7 +114,7 @@ public HttpResponseMessage CreateUser(CreateUserContract contract)
                     PageSize = pageSize,
                     SortColumn = sortColumn,
                     SortAscending = sortAscending,
-                    PortalId = PortalId,
+                    PortalId = PortalController.GetEffectivePortalId(PortalId),
```
kanm2015: do we need this change in other user related API methods?

kanm2015:
code review done

PR #343, opened 2017-10-25
DNN-10403: fixed dup tags issue on page settings

kanm2015:


amarjit-dhunna (commented):
@kanm2015 , The code behind changes look good but I think we should filter the tags on the UI as well. In that case input UI won't mismatch with the actual saved one.

PR #352, opened 2017-10-30
DNN-10414: prevent from selecting current page itself or its descenda…

kanm2015:
…nt as the parent


zyhfish (commented):

PR #355, opened 2017-11-01
CONTENT-8581: set modules to null when add new page, so it can copy all modules from template page if defined.

zyhfish:


kanm2015:
code review done

PR #360, opened 2017-11-04
DNN-10425: fixing duplicate results on first page of Lucene's users search.

galatrash:


src/Modules/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
```diff
@@ -502,12 +502,14 @@ private static IList<UserBasicDto> GetUsersFromLucene(GetUsersContract usersCont
...
                 return userId;
-            }).Where(u => u > 0);
+            }).Where(u => u > 0).Distinct().ToList();
 
             var currentIds = string.Join(",", userIds);
             var users = UsersDataService.Instance.GetUsersByUserIds(usersContract.PortalId, currentIds)
                 .Where(u => UserController.GetUserById(usersContract.PortalId, u.UserId).Membership.Approved).ToList();
             totalRecords = searchResults.TotalHits;
+            if (usersContract.PageIndex == 0 && query.PageSize > userIds.Count)
+                totalRecords -= query.PageSize - userIds.Count;
```
kanm2015, line 512: If I have totalRecords = 15, userIds.Count = 6, pageIndex = 0, pageSize = 10, the updated totalRecords will be 11? I will then see 6 users on the first page and total pages will be 2, which doesn't make sense.

galatrash: With Lucene, it can be impossible to know the exact number unless we retrieve all the results and filter unique records. This will make it take us back to the ame situation before starting on this issue. Instead of keeping dragging this on, IMO, we should go back to database search for exact search results.

kanm2015: +1, only use db search for users here

kanm2015:
close this one since problem will be solved in a different way

PR #362, opened 2017-11-06
DNN-10455: updated to use mouse enter/leave events for page context menu

kanm2015:


amarjit-dhunna (commented):
@kanm2015  Changes are good. Please add animations for the flyout menu

PR #363, opened 2017-11-07
DNN-9112: PB SQL Console: Allows XSS scripts

Mohtshum:
- Did the decoding before notify method.
- Remove the newline replacement after verified that has not been used by bad sql queries.


kanm2015:
code review done

PR #364, opened 2017-11-07
DNN-10378: Fixed the permissions when creating a page.

amarjit-dhunna:
Validate the page settings for all types.
Add the page to list even if not normal type.


kanm2015:
code review done

PR #365, opened 2017-11-07
DNN-9985: prevent from adding page url to list if url linking to non-…

kanm2015:
…existing portal alias


zyhfish:

PR #366, opened 2017-11-09
CONTENT-8590: force to reload site list on site settings if accessing…

kanm2015:
… via sites


zyhfish (approved):

PR #370, opened 2017-11-20
DNN-10215: update to allow comma separated emails

kanm2015:


src/Modules/Manage/Dnn.PersonaBar.AdminLogs/AdminLogs.Web/src/components/AdminLog/EmailPanel/index.jsx
```diff
@@ -74,7 +74,7 @@ class EmailPanel extends Component {
 
     validateEmail(value) {
         const re = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-        return re.test(value);
+        return value.split(",").every(v => re.test(v));
```
galatrash: Shouldn't we also allow semicolons as well? Many email clients send with semicolon as a separator between multiple emails. Also, shouldn't you trim leading/trailing whitespace from the results?

kanm2015: thx, will take this into account

PR #371, opened 2017-11-21
DNN-10583: Support for portal id in web api method

donker:
Currently the WebAPI endpoint is protected to allow host users only. It makes sense that the endpoint will allow hosts to run commands against other portals then the one currently being used. This will help us create more powerful possibilities in the future with the powershell connector.

### DIFF
--- a/src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
+++ b/src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
@@ -27,36 +27,36 @@ namespace Dnn.PersonaBar.Prompt.Services
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(CommandController));
         private static readonly string[] BlackList = { "smtppassword", "password", "pwd", "pass", "apikey" };
 
-        private int portalId = -1;
+        private int _portalId = -1;
         private new int PortalId
         {
             get
             {
-                if (portalId == -1)
+                if (_portalId == -1)
                 {
-                    portalId = base.PortalId;
+                    _portalId = base.PortalId;
                 }
-                return portalId;
+                return _portalId;
             }
             set
             {
-                portalId = value;
+                _portalId = value;
             }
         }
-        private PortalSettings portalSettings;
+        private PortalSettings _portalSettings;
         private new PortalSettings PortalSettings
         {
             get
             {
-                if (portalSettings == null)
+                if (_portalSettings == null)
                 {
-                    portalSettings = base.PortalSettings;
+                    _portalSettings = base.PortalSettings;
                 }
-                return portalSettings;
+                return _portalSettings;
             }
             set
             {
-                portalSettings = value;
+                _portalSettings = value;
             }
         }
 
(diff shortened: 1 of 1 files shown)


david-poindexter:
Excellent @donker !

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
```diff
@@ -16,20 +16,61 @@
...
 using DotNetNuke.Services.Localization;
 using DotNetNuke.Services.Log.EventLog;
+using DotNetNuke.Entities.Portals;
 
 namespace Dnn.PersonaBar.Prompt.Services
 {
     [MenuPermission(MenuName = "Dnn.Prompt")]
     [RequireHost]
-    public class CommandController : ControllerBase
+    public class CommandController : ControllerBase, IServiceRouteMapper
```
galatrash-at-dnn: Why this is added? The PB services don't need it.

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
```diff
@@ -16,20 +16,61 @@
...
     {
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(CommandController));
         private static readonly string[] BlackList = { "smtppassword", "password", "pwd", "pass", "apikey" };
 
-        [HttpGet]
-        public HttpResponseMessage List()
```
galatrash, line 30: Why this method has been removed?

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
```diff
@@ -16,20 +16,61 @@
...
+        private int portalId = -1;
```
galatrash: I suggest naming `portalId` to `_portalId` to differentiate from passed parameter and to follow conventions.

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
```diff
@@ -16,20 +16,61 @@
...
+                    portalId = base.PortalId;
+                }
+                return portalId;
+            }
+            set
+            {
+                portalId = value;
+            }
+        }
+        private PortalSettings portalSettings;
```
galatrash: I suggest naming `portalSettings` to `_portalSettings` to differentiate from passed parameter and to follow conventions.

donker: Sure. Feel free. Or do I need to do that?

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
donker: The "List" method was never used and superceded by the List command that passed through the main WebAPI method. I removed it to remove an unused endpoint.

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
galatrash: We have a convention that the person who reviews shouldn't do any changes in the code and leave it to the PR submitter. Therefore, I prefer that you do these changes and I will review and approve.

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
mean2me: Hi guys, list method was used by "HELP" command and actually removing it was a breaking change that caused DNN-10598

donker: OK, I'll look into this. We'll do more testing in the Prompt group here. We're making sure we're keeping endpoints to the bare minimum.

PR #373, opened 2017-11-24
DNN-10425: fixing PB get user with filtering performance.

galatrash:
Also (wip) started adding integration tests.

Note: I couldn't continue the tests due to another bug in the Platform - DNN-10547.

### DIFF
--- a/src/Modules/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
+++ b/src/Modules/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
@@ -32,7 +32,6 @@ using System.Threading;
 using System.Web.Security;
 using Dnn.PersonaBar.Users.Components.Contracts;
 using Dnn.PersonaBar.Users.Components.Dto;
-using Dnn.PersonaBar.Users.Data;
 using DotNetNuke.Common.Utilities;
 using DotNetNuke.Data;
 using DotNetNuke.Entities.Users;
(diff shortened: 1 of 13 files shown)


src/Modules/Manage/Dnn.PersonaBar.Users/Providers/DataProviders/SqlDataProvider/Uninstall.SqlDataProvider
```diff
@@ -16,6 +16,9 @@ IF EXISTS (SELECT * FROM dbo.sysobjects where id = object_id(N'{databaseOwner}{o
 	DROP PROCEDURE {databaseOwner}{objectQualifier}Personabar_GetUsers
 GO
 
+IF EXISTS (SELECT * FROM sys.procedures WHERE object_id = object_id(N'{databaseOwner}[{objectQualifier}Personabar_GetUsersBySearchTerm]'))
+	DROP PROCEDURE {databaseOwner}[{objectQualifier}GetUsersBySearchterm]
```
sleupold: should be 
DROP PROCEDURE {databaseOwner}[{objectQualifier}Personabar_GetUsersBySearchterm]

src/Modules/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
```diff
@@ -390,65 +381,52 @@ public static bool IsAdmin(PortalSettings portalSettings)
...
                     break;
                 case UserFilters.Deleted:
-                    users = GetUsers(usersContract, true, true, isSuperUser, out totalRecords);
-                    paged = true;
+                    users = GetUsers(usersContract, null, true, isSuperUser ? null : (bool?)false, out totalRecords);
                     break;
-                //                    case UserFilters.Online:
-                //                        dbUsers = UserController.GetOnlineUsers(usersContract.PortalId);
-                //                        break;
                 case UserFilters.RegisteredUsers:
```
sleupold, line 410: What is the difference between "Authorized" state and member of "Registered Users" role - AFAIK there is none.
The only interesting role I, could think of, would be Administrators role - but there is another option to view its members.

galatrash: THe registered user is a role while the authorized is a state; they are totally independent. I don't know what did it mean in the first place to have users in the Registered Users role; but this is a history and goes way before I got involved with DNN :)

sleupold: AFAIK membership in registeredUsers role is maintained automatically by DNN, when a user gets authorized and unauthorized or deleted

kanm2015:
code review done

PR #383, opened 2017-12-15
Fix to prompt script to eliminate access to deprecated List WebAPI

donker:
I removed the List endpoint in PR 371. There was one prompt call that used this: "help". It does a call to "list-commands" but returns it as a flat list instead of a full prompt return object. So all it required was removal of the method from the service js and rewrite the call to use the regular cmd call.


mean2me (approved):
It solves API call problem.

PR #387, opened 2017-12-17
DNN-10156 - Only render a component if it will be visible

ricardodiasc:
Avoid not necessary render and server request when a tab is not visible.


kanm2015:
code review done

PR #395, opened 2017-12-26
DNN-10468 - Fix the presentation of page parent when adding or editing a page.

ricardodiasc:


mean2me (approved):
Ok for me.

tpluscode:
Disregard the previous commit

PR #400, opened 2017-12-28
DNN-10634 Fixed

mean2me:


src/Modules/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/src/store/configureStore.js
```diff
@@ -3,15 +3,18 @@ import thunkMiddleware from "redux-thunk";
...
+    if (IS_DEV) {
+        return createStore(
+            reducers,
+            initialState,
+            compose(applyMiddleware(thunkMiddleware, reduxImmutableStateInvariant()), DevTools.instrument()));
+    } else {
+        return createStore(
+            reducers,
+            initialState,
+            compose(applyMiddleware(thunkMiddleware), DevTools.instrument()));
```
tpluscode: Hm, maybe `DevTools` should also be excluded from production?

PR #403, opened 2017-12-29
DNN-10470 File upload placeholder string has been fixed.

mean2me:


tpluscode:
Typo is the only necessary change? What happened to the yarn.lock file?

mean2me:
About yarn.lock, I committed accidentally. Gonna fix it.

PR #413, opened 2018-01-09
DNN-10481 

mean2me:
Fix: Edit menu item not visible for page types: Existing, url and File


src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/dnn-persona-bar-page-treeview/src/_PersonaBarTreeInContextMenu.jsx
```diff
@@ -57,37 +50,81 @@ export class PersonaBarTreeInContextMenu extends Component {
                 break;
         }
     }
-    sort(items, column, order) {
```
tpluscode, line 60: Having removed the `sort` method, are you sure it isn't called anywhere that this component is used?

mean2me: yes, used only once

PR #415, opened 2018-01-10
DNN-10663 Fixed missing OK messages.

mean2me:


src/Modules/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/src/components/Output.jsx
```diff
@@ -33,15 +33,14 @@ export class Output extends Component {
...
             if(props.paging !== null) {
                 props.IsPaging(props.paging.pageNo < props.paging.totalPages);
-                const result = [];
-                result.push(this.renderData(props.data, fieldOrder));
-                result.push(<TextLine key={DomKey.get("output")} txt={props.output} css="dnn-prompt-ok"/>);
-                return result;
-            } else {
-                return this.renderData(props.data, fieldOrder);
             }
+            result.push(this.renderData(props.data, fieldOrder));
```
tpluscode, line 40: Maybe these should be wrapped in an `else` if both paging and output are present? Unless it is desired to show both in such case?

mean2me: when empty, "output" is an empty string and seems to be useful for spacing with current layout

PR #417, opened 2018-01-11
DNN-10485 Optimized word truncation logic

mean2me:


tpluscode:
Could we use CSS `ellipsis` rule instead?

mean2me:
It would be the best approach, as I commented on JIRA, but the way the component is designed doesn't work with a responsive design. We should do some refactoring to make it a responsive component.

PR #419, opened 2018-01-11
DNN-10678: hide the add page & add multiple pages buttons in search r…

kanm2015:
…esult view and display confirmation before switching to search from add page(s) and add template view


zyhfish (approved):

zyhfish:

PR #420, opened 2018-01-12
DNN-10661: validate module whether is null.

zyhfish:


src/Modules/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/src/components/TextLine.jsx
```diff
@@ -3,7 +3,10 @@ import Parser from "html-react-parser";
 import DomKey from "../services/DomKey";
 
 const TextLine = ({txt, css}) => {
-    const textLines = txt.split("\n");
+    if(!txt){
+        return null;
```
tpluscode, line 7: Can returning null potentially break in another place?

zyhfish: nope, null means nothing to render.

PR #423, opened 2018-01-12
DNN-10683: added null check

kanm2015:


zyhfish:

zyhfish (approved):

PR #427, opened 2018-01-16
DNN-10477: Published Date filter wont work for current date

Mohtshum:
- Initially, add the last hour in current date for ENDDATE
- Add time in comparision of start and end date


kanm2015:
please fix it after DNN-10659 is done, thx

PR #430, opened 2018-01-17
DNN-10702: fixed table styling issue when there's long txt in column

kanm2015:


zyhfish (approved):

PR #433, opened 2018-01-19
DNN-10696: Prompt: "get-module" command return false data in case of …

Mohtshum:
…inappropriate page id

- Add check if module not found for pageid


kanm2015:
code review done

PR #443, opened 2018-01-23
DNN-10330 Fixed broken "list-users --role ROLE_NAME"

mean2me:
Description: there was an error in pagination of the result. It started from 1 as a default page index, intead of 0


tpluscode:
Oh, if default page was `1`, then it must have affected all other usages of `list-users` command, right?

PR #445, opened 2018-01-24
DNN-10337 Fixed wrong condition not allowing user deletion.

mean2me:
Added missing messages in referenced .resx file.

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Recyclebin/Components/Prompt/Commands/PurgeUser.cs
+++ b/src/Modules/Content/Dnn.PersonaBar.Recyclebin/Components/Prompt/Commands/PurgeUser.cs
@@ -28,7 +28,7 @@ namespace Dnn.PersonaBar.Recyclebin.Components.Prompt.Commands
             var userInfo = UserController.Instance.GetUser(PortalId, UserId);
             if (userInfo == null)
                 return new ConsoleErrorResultModel(string.Format(LocalizeString("UserNotFound"), UserId));
-            if (!userInfo.IsDeleted)
+            if (userInfo.IsDeleted)
                 return new ConsoleErrorResultModel(LocalizeString("Prompt_CannotPurgeUser"));
 
             RecyclebinController.Instance.DeleteUsers(new List<UserInfo> { userInfo });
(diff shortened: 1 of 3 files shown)


src/Modules/Content/Dnn.PersonaBar.Recyclebin/Components/Prompt/Commands/PurgeUser.cs
```diff
@@ -28,7 +28,7 @@ public override ConsoleResultModel Run()
             var userInfo = UserController.Instance.GetUser(PortalId, UserId);
             if (userInfo == null)
                 return new ConsoleErrorResultModel(string.Format(LocalizeString("UserNotFound"), UserId));
-            if (!userInfo.IsDeleted)
+            if (userInfo.IsDeleted)
```
kanm2015: only purge soft deleted user, should not change here

src/Modules/Content/Dnn.PersonaBar.Recyclebin/Components/RecyclebinController.cs
```diff
@@ -332,7 +332,7 @@ public void DeleteUsers(IEnumerable<UserInfo> users)
                 var user in
                     userInfos.Select(u => UserController.GetUserById(u.PortalID, u.UserID))
                         .Where(user => user != null)
-                        .Where(user => user.IsDeleted))
+                        .Where(user => !user.IsDeleted))
```
kanm2015: only purge soft deleted user, should not change here

src/Modules/Content/Dnn.PersonaBar.Recyclebin/Components/Prompt/Commands/PurgeUser.cs
mean2me: What's the correct logic? As it was before, only soft-deleted users will be purged. With this new logic, if a user has been softly deleted (for some reason) I don't purge it.

src/Modules/Content/Dnn.PersonaBar.Recyclebin/Components/RecyclebinController.cs
mean2me: Same as previous comment.

kanm2015:
code review done

PR #446, opened 2018-01-24
DNN-10324 Fixed issue in set-role command

mean2me:


src/Modules/Manage/Dnn.PersonaBar.Roles/Components/Prompt/Commands/SetRole.cs
```diff
@@ -66,7 +66,7 @@ public override void Init(string[] args, PortalSettings portalSettings, UserInfo
                     break;
             }
 
-            if (string.IsNullOrEmpty(RoleName) && string.IsNullOrEmpty(Description) && !IsPublic.HasValue && !AutoAssign.HasValue)
+            if ((string.IsNullOrEmpty(RoleName) && RoleId < 0) && string.IsNullOrEmpty(Description) && !IsPublic.HasValue && !AutoAssign.HasValue && string.IsNullOrEmpty(status))
```
kanm2015: why check "status" here, seems not needed

mean2me: Hi Kan, I added this check to cover case like "set-role ROLE_ID --status Pending".
If I don't check for "status", that case will return an error.

kanm2015: status is not a mandatory param, it should still work if we missing it. I tested it a little bit and seems no error returned as long as I passed valid user Id. could you take a look again? thx

kanm2015: and for the case status not provided, the role will just stay on its current status

PR #448, opened 2018-01-25
DNN-10633: stop triggering culture code change event within validateC…

kanm2015:
…ultureCode


kanm2015:
it causes duplicate web api calls, so get rid of this event trigger.

PR #449, opened 2018-01-25
DNN-10709: Prompt Cross-portal scripting: Modules commands work regar…

Mohtshum:
…dless the supplied portalID and moves/copies modules even while they are not in a site group

- Add portalsetting in Get & Delete module, to evaluate the site group and compare cross portal issue


kanm2015:
code review done

PR #452, opened 2018-01-27
DNN-10705: Prompt: Cross-portal scripting: "new-page" always creating…

Mohtshum:
… pages in the main site

- Passing the portalSettings to utilize the site group
- API calls already allowing only site group for crossportal DNN-10709


kanm2015:
code review done

PR #461, opened 2018-02-01
Dnn-10735: Prompt: In case of across-portal scripting, sending invali…

Mohtshum:
…d portal Id in the API, responds with a message related to site groups which has nothing to do with the invalid portalID

- Check for not existing portal


kanm2015:
code review done

PR #463, opened 2018-02-02
Dnn-10689 : Prompt: Spacing between command and previous/next results…

Mohtshum:
… got removed, commands appear to be stacked

- put a rule to give top & bottom margin for space of commands executed


src/Modules/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/src/components/Prompt.less
```diff
@@ -37,6 +37,9 @@
 
                 .dnn-prompt-output {
                     /*pending*/
```
ricardodiasc: This code is pending something?

ricardodiasc (approved):

PR #469, opened 2018-02-07
DNN-10358 - Calculate the treeview space and changes on how dnn-scrol…

ricardodiasc:
…lbar is imported


kanm2015:
code review done

PR #472, opened 2018-02-07
DNN-10709 Prompt: Cross-portal scripting: Modules commands work regar…

Mohtshum:
…dless the supplied portalID and moves/copies modules even while they are not in a site group

- Only allowing get-module commands if without portalId for cross-scripting

### DIFF
--- a/src/Modules/Settings/Dnn.PersonaBar.Prompt/Components/ModulesController.cs
+++ b/src/Modules/Settings/Dnn.PersonaBar.Prompt/Components/ModulesController.cs
@@ -113,7 +113,13 @@ namespace Dnn.PersonaBar.Prompt.Components
                 return null;
             }
 
-            if (targetPage.PortalID == portalSettings.PortalId  || PortalHelper.IsRequestForSiteGroup(targetPage.PortalID,portalSettings.PortalId))
+            var currentPortalSetting = PortalController.Instance.GetCurrentPortalSettings();
+
+            if (
+                currentPortalSetting == portalSettings ||
+                targetPage.PortalID == portalSettings.PortalId ||
+                PortalHelper.IsRequestForSiteGroup(targetPage.PortalID, portalSettings.PortalId)
+                )
             {
                 try
                 {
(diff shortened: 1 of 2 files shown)


src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
```diff
@@ -65,24 +65,17 @@ private new PortalSettings PortalSettings
         public HttpResponseMessage Cmd(int portalId, [FromBody] CommandInputModel command)
         {
             PortalId = portalId;
-            var portal = PortalController.Instance.GetPortal(PortalId);
+            var portal = PortalController.Instance.GetPortalSettings(portalId);
```
kanm2015: why we need this change?

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
```diff
@@ -65,24 +65,17 @@ private new PortalSettings PortalSettings
...
-            {
-                if (!PortalHelper.IsRequestForSiteGroup(portalId, base.PortalSettings.PortalId))
-                {
-                    var errorMessage = string.Format(Localization.GetString("InValidPortalId", Constants.LocalResourcesFile), portalId);
-                    Logger.Error(errorMessage);
-                    return AddLogAndReturnResponse(null, null, command, DateTime.Now, errorMessage);
-                }
-                PortalSettings = new PortalSettings(PortalId);
-            }
+            PortalSettings = new PortalSettings(PortalId);
```
kanm2015, line 77: portal id check not taken in account any more?

src/Modules/Settings/Dnn.PersonaBar.Prompt/Components/ModulesController.cs
```diff
@@ -113,7 +113,13 @@ public ModuleInfo CopyModule(PortalSettings portalSettings, int moduleId, int so
                 return null;
             }
 
-            if (targetPage.PortalID == portalSettings.PortalId  || PortalHelper.IsRequestForSiteGroup(targetPage.PortalID,portalSettings.PortalId))
+            var currentPortalSetting = PortalController.Instance.GetCurrentPortalSettings();
+
+            if (
+                currentPortalSetting == portalSettings ||
```
kanm2015, line 119: can we compare portal id instead of settings here?

src/Modules/Settings/Dnn.PersonaBar.Prompt/Services/CommandController.cs
Mohtshum: This was much needed to differentiate in rest of life cycle that either we have different PortalSettings object than the settings obtained from PortalController.Instance.GetCurrentPortalSettings() which is by default for request.

src/Modules/Settings/Dnn.PersonaBar.Prompt/Components/ModulesController.cs
Mohtshum: No way, as I commented above that was much needed did enable us to differentiate a request for portalId either passed for cross-scripting or not. And it really helped to understand if for example, host user has token for site3 but not passing the crossportalid, and trying to access module for site1, which he should get and this check enabling him and if same time he passed the portal id 3 then he should not get.

kanm2015:
changes look good, merged

PR #479, opened 2018-02-10
dDNN-10681: Prompt: Across-portal scripting: "list-pages" always listi…

Mohtshum:
…ng pages for the main site even with a remote site id Or with the portal-Id specified not for an existing portal

- Passed portalSettings to return data against it


kanm2015:
code review done

PR #486, opened 2018-02-16
DNN-8325

mean2me:
…ub-pages sorted by ID, PARENT_ID

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
+++ b/src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
@@ -322,6 +322,70 @@ namespace Dnn.PersonaBar.Pages.Components
             return pages;
         }
 
+        private IEnumerable<TabInfo> GetPageListHelper(string searchKey = "", bool includeHidden = true, bool includeDeleted = false)
+        {
+            var portalSettings = PortalController.Instance.GetCurrentPortalSettings();
+            var adminTabId = portalSettings.AdminTabId;
+
+            var tabs = TabController.GetPortalTabs(portalSettings.PortalId, adminTabId, false, includeHidden, includeDeleted, true);
+            var pages = from t in tabs
+                        where (t.ParentId != adminTabId || t.ParentId == Null.NullInteger) &&
+                                !t.IsSystem &&
+                                    (string.IsNullOrEmpty(searchKey)
+                                        || (!string.IsNullOrEmpty(searchKey) &&
+                                                (t.TabName.IndexOf(searchKey, StringComparison.OrdinalIgnoreCase) > Null.NullInteger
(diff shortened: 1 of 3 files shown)


src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -322,6 +322,70 @@ public IEnumerable<TabInfo> GetPageList(int parentId = -1, string searchKey = ""
...
+
+            var sorted = pages.OrderBy(x => x.ParentId > -1 ? x.ParentId : x.TabID).ThenBy(x => x.TabID);
+            return sorted;
+        }
+
+        public IEnumerable<TabInfo> GetSubPagesList(bool? deleted, string tabName, string tabTitle, string tabPath, string tabSkin, bool? visible, out int total, string searchKey = "", int pageIndex = -1, int pageSize = 10)
+        {
+            pageIndex = pageIndex <= 0 ? 0 : pageIndex;
+            pageSize = pageSize > 0 && pageSize <= 100 ? pageSize : 10;
+            var tabs = GetPageListHelper(searchKey, true, deleted ?? false);
```
tpluscode: This line is the only difference between `GetSubPagesList` and `GetPageList` and this entire methods is otherwise an exact copy. This is no good. 

There should be a single method with all the logic below. Maybe accepting the `tabs` as an additional param?

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -322,6 +322,70 @@ public IEnumerable<TabInfo> GetPageList(int parentId = -1, string searchKey = ""
             return pages;
         }
 
+        private IEnumerable<TabInfo> GetPageListHelper(string searchKey = "", bool includeHidden = true, bool includeDeleted = false)
```
tpluscode: This method is very similar to `public IEnumerable<TabInfo> GetPageList`. I see two differences: 

1. no `(t.ParentId == parentId)` check
1. added sorting

The first is fine but what about the second? Why add sorting where the other does not do ordering?

Would you try to extract a common denominator?

tpluscode: Also, not about your code per se, but the multi-line `where` statement is mighty awful. 

Please refactor this. I'd start with extracting parts of that expression into [`let` statements](https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/let-clause)

tpluscode (changes requested):

mean2me:
Code reviewed

tpluscode (approved):

tpluscode:
Much better, thank you!

PR #487, opened 2018-02-18
DNN-10760 : Pages "All Users" permissions are not saved

Mohtshum:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -1112,7 +1112,7 @@ public void SavePagePermissions(TabInfo tab, PagePermissions permissions)
                 foreach (var rolePermission in permissions.RolePermissions.Where(NoLocked()))
                 {
                     var role = RoleController.Instance.GetRoleById(portalSettings.PortalId, rolePermission.RoleId);
-                    if (role != null)
+                    if (role != null || (rolePermission.RoleId.ToString() == Globals.glbRoleAllUsers))
```
tpluscode: Call it micro-optimisation, but maybe we could first check the `RoleId` and skip calling `RoleController.Instance.GetRoleById` when it's "All Users"

Mohtshum: A perfect example for Short circuit. Unbeliveable that I'missed!

Mohtshum: Changes made thnx

PR #492, opened 2018-03-04
DNN-16884: Fix Page Parent style when parent page name has no spaces …

ricardodiasc:
…and it is too big

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageTypeSelector/PageTypeSelector.jsx
+++ b/src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageTypeSelector/PageTypeSelector.jsx
@@ -5,6 +5,9 @@ import Localization from "../../localization";
 import RadioButtons from "dnn-radio-buttons";
 import utils from "../../utils";
 
+
+const MAX_PAGE_PARENT_SIZE_TOOGLE = 180;
+
 class PageTypeSelector extends Component {
 
     constructor(props) {
@@ -33,17 +36,16 @@ class PageTypeSelector extends Component {
 
     componentDidMount() {
         const pageSelectorDOM = ReactDOM.findDOMNode(this);
-        if (pageSelectorDOM.querySelector("#parentPageValue").offsetWidth > 180) {
+        if (pageSelectorDOM.querySelector("#parentPageValue").offsetWidth > MAX_PAGE_PARENT_SIZE_TOOGLE) {
             pageSelectorDOM.querySelector("#pageParent").className = "page-info-item page-parent-info-style-large";
             pageSelectorDOM.querySelector("#pageParentLabel").className = "page-info-item-label parent-page-style-label-large";
             pageSelectorDOM.querySelector("#pageParentContent").className = "page-info-item-value parent-page-name parent-page-style-content-large";
         }
-        console.log("Component did mount",pageSelectorDOM.querySelector("#parentPageValue").offsetWidth);
     }
 
     _calculateParentPageSize(parentPageRef) {
         if (parentPageRef !== null) {
-            this.pageParentLarge = (parentPageRef.offsetWidth > 180);
+            this.pageParentLarge = (parentPageRef.offsetWidth > MAX_PAGE_PARENT_SIZE_TOOGLE);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageTypeSelector/PageTypeSelector.jsx
```diff
@@ -33,6 +33,32 @@ class PageTypeSelector extends Component {
         const hierarchy = this._getHierarchyLabel();        
         const components = this.getComponents(); 
 
+        let pageParentInfoStyle = null;
+        let parentPageStyleLabel = null;
+        let parentPageStyleContent = null;
+        
+
+        if (hierarchy.length < 31) {
```
tpluscode: Why did you choose 31 as the cutoff length? Keeping in mind that the font is variable width, is it low enough to give a desired effect when the letters are wider on average?

src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageTypeSelector/PageTypeSelector.jsx
```diff
@@ -51,13 +77,13 @@ class PageTypeSelector extends Component {
                                 {createdDate}
                             </span>
                         </div>
-                        <div className="page-info-item">
-                            <span className="page-info-item-label">
+                        <div className="page-info-item" style={pageParentInfoStyle}>
```
tpluscode: I'd rather we avoid inline styles and instead use class or attribute selectors. Is there some problem with selector specificity which blocks us?

What I have in mind is to set a class based on the logic above

```
<div className="page-info-item {hierarchy.length > 31 ? page-info-long : null}" />
```

And style those divs purely in css. Thoughts @emac80 ?

src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageTypeSelector/PageTypeSelector.jsx
ricardodiasc: Agree on that. I will change to find out the width if this field

src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageTypeSelector/PageTypeSelector.jsx
```diff
@@ -24,41 +30,33 @@ class PageTypeSelector extends Component {
...
     }
+
+    componentDidMount() {
+        const pageSelectorDOM = ReactDOM.findDOMNode(this);
+        if (pageSelectorDOM.querySelector("#parentPageValue").offsetWidth > 180) {
+            pageSelectorDOM.querySelector("#pageParent").className = "page-info-item page-parent-info-style-large";
+            pageSelectorDOM.querySelector("#pageParentLabel").className = "page-info-item-label parent-page-style-label-large";
+            pageSelectorDOM.querySelector("#pageParentContent").className = "page-info-item-value parent-page-name parent-page-style-content-large";
+        }
+        console.log("Component did mount",pageSelectorDOM.querySelector("#parentPageValue").offsetWidth);
```
tpluscode: Looks like a debugging log

src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageTypeSelector/PageTypeSelector.jsx
```diff
@@ -24,41 +30,33 @@ class PageTypeSelector extends Component {
             return page.hierarchy;
         }
     }
+
+    componentDidMount() {
+        const pageSelectorDOM = ReactDOM.findDOMNode(this);
+        if (pageSelectorDOM.querySelector("#parentPageValue").offsetWidth > 180) {
```
tpluscode: Let's have this magical 180 extracted into a constant and we'll be done here!

tpluscode (approved):
Nicely done. Using actual element dimensions is so much better! 💖 

Please address the last two nitpicks and I'll be happy to merge.

PR #494, opened 2018-03-07
DNN-17650: add password question/answer field.

zyhfish:


src/Modules/Manage/Dnn.PersonaBar.Users/Users.Web/src/_exportables/src/components/CreateUserBox/index.jsx
```diff
@@ -203,7 +227,7 @@ class CreateUserBox extends Component {
                                 tooltipMessage={Localization.get("Password.Help")}
                                 errorMessage={Localization.get("Password.Required") }
                                 style={inputStyle}
-                                inputStyle={{ marginBottom: 15 }}
+                                inputStyle={!requiresQuestionAndAnswer ? { marginBottom: 15 } : { marginBottom: 0 }}
```
tpluscode, line 230: I see that this and other `SingleLineInputWithError` elements here use inline styles. Can we extract them to a stylesheet and decorate with classes?

zyhfish: The common component of SingleLineInputWithError used inline style, so we need use inline style as well to overwrite the default style.

PR #495, opened 2018-03-07
DNN-8330

mean2me:


src/Modules/Manage/Dnn.PersonaBar.Roles/Components/Prompt/Commands/SetRole.cs
```diff
@@ -76,7 +76,12 @@ public override ConsoleResultModel Run()
         {
             try
             {
+
                 var existingRole = RoleController.Instance.GetRoleById(PortalId, RoleId);
+                if (existingRole.IsSystemRole)
```
tpluscode, line 82: Isn't this case handled inside the `RolesController`?

src/Modules/Manage/Dnn.PersonaBar.Roles/Components/Prompt/Commands/SetRole.cs
```diff
@@ -101,8 +106,8 @@ public override ConsoleResultModel Run()
             }
             catch (Exception ex)
             {
-                Logger.Error(ex);
-                return new ConsoleErrorResultModel(LocalizeString("RoleUpdated.Error"));
+                Logger.Error(ex);
+                return new ConsoleErrorResultModel(LocalizeString(!String.IsNullOrEmpty(ex.Message) ? ex.Message : "RoleUpdated.Error"));
```
tpluscode: `ex.Message` can contain very technical information. Not sure if it's the best thing to show directly to the user. At least on its own.

By the way, using it on the `LocalizeString` with any arbitrary string will have no effect. Please refer to [this function](https://github.com/dnnsoftware/Dnn.Platform/blob/development/DNN%20Platform/Library/Services/Localization/LocalizationProvider.cs#L72) which will be called eventually

PR #498, opened 2018-03-14
DNN-11748 Cannot Copy Page with specific user permissions

Mohtshum:
- Changes for parentId
- Setting up page picker to parent page

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/services/pageService.js
+++ b/src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/services/pageService.js
@@ -18,21 +18,21 @@ const PageService = function () {
     };
 
     const savePage = function (page) {
-        const api = getOverridablePagesApi();
-        let request = page;
+        const api = getOverridablePagesApi();        
+        let additionalParams = null;
 
         if (page.tabId === 0 && !securityService.isSuperUser()) {
-            request = {
-                ...request,
+            additionalParams = {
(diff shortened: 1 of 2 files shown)


src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/services/pageService.js
```diff
@@ -21,10 +21,18 @@ const PageService = function () {
...
+                ...request,
+                parentId: page.parentId !== undefined && page.parentId !== null ? utils.getCurrentPageId() : page.parentId
+            };
+        }
+        else {
+            request = {
+                ...request,
+                url: ""
+            };
+        }
```
tpluscode: I think this fragment is kind of verbose. I'd simplify and reduce nesting by extracting the `parentId`/`url` into a variable and inline `page` to have a single assignment

```js
var request = { ...page, ...additionalParams };
```

tpluscode (changes requested):
There any back-end changes necessary, right?

src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/services/pageService.js
Mohtshum: No we don't need any backend changes, just need to give all permissions to the user, whom we want to duplicate the page

tpluscode (approved):

PR #500, opened 2018-03-15
DNN-8337

mean2me:


mean2me (commented):
Some files had a mix of LF/CRLF EOL. Set all to LF.

src/Modules/Manage/Dnn.PersonaBar.Roles/Components/IRolesController.cs
```diff
@@ -10,6 +10,7 @@ public interface IRolesController
     {
         bool SaveRole(PortalSettings portalSettings, RoleDto roleDto, bool assignExistUsers, out KeyValuePair<HttpStatusCode, string> message);
         IEnumerable<RoleInfo> GetRoles(PortalSettings portalSettings, int groupId, string keyword, out int total, int startIndex, int pageSize);
+        List<RoleInfo> GetRolesByNames(PortalSettings portalSettings, int groupId, string commaSeparatedRoles, out List<string> notFound);
```
tpluscode: About the signature:

1. Passing a comma separated string smells of violated SRP. Please change the parameter to an array/list and move splitting to calling code
1. Please return interfaces instead of concrete implementations. `IList<T>` or `IEnumerable<T>` in this case.
1. `out` parameters are frowned upon outside `TryParse`-like methods. In general were introduced as COM compatibility. Unfortunately C# doesn't (yet) support [anonymous return types][value tuple]. I come up with 4 options:
   1. dedicated class/struct to return, combining both found and not found roles as two fields
   1. strongly typed `Tuple<IList<RoleInfo>, List<string>>` - kind of icky on a public API, because has a weak API
   1. `IDictionary<string, RoleInfo>` - unconventional maybe and also suffers from weak semantics
   1. *my favourite* - just return the found roles and let the caller handle identifying that some/which are missing. this method probably has enough responsibility already

[value tuple]: https://blogs.msdn.microsoft.com/mazhou/2017/05/26/c-7-series-part-1-value-tuples/

src/Modules/Manage/Dnn.PersonaBar.Roles/Components/RolesController.cs
```diff
@@ -64,6 +72,50 @@ public IEnumerable<RoleInfo> GetRoles(PortalSettings portalSettings, int groupId
...
+                    }
+                    return false;
+                }).ToList();
+            
+            }
+
+            rolesList.ExceptWith(foundNames);
+            notFound = rolesList.ToList();
+
+            return foundRoles;
```
tpluscode: When `commaSeparatedRoles` is "`NullOrWhiteSpace`" then this method will also return null. When possible please avoid returning null.

src/Modules/Manage/Dnn.PersonaBar.Users/Components/Prompt/Commands/AddRoles.cs
```diff
@@ -1,70 +1,87 @@
-﻿using System;
```
tpluscode, line 1: I appreciate fixing line endings but it should be done in a separate PR. Now I cannot tell what are the actual code changes 🔍

src/Modules/Manage/Dnn.PersonaBar.Roles/Components/RolesController.cs
```diff
@@ -64,6 +72,50 @@ public IEnumerable<RoleInfo> GetRoles(PortalSettings portalSettings, int groupId
...
+            HashSet<string> foundNames = new HashSet<string>();
+            if (rolesList.Count() > 0)
+            {
+                var allRoles = (groupId < Null.NullInteger
+                ? RoleController.Instance.GetRoles(portalSettings.PortalId)
+                : RoleController.Instance.GetRoles(portalSettings.PortalId, r => r.RoleGroupID == groupId));
+                
+                foundRoles = allRoles.Where(r => 
+                {
+                    bool adminCheck = isAdmin || r.RoleID != portalSettings.AdministratorRoleId;
```
tpluscode, line 96: Mm, I won't question this logic in this particular but I see it occurs in a number of methods in this class. Maybe we could extract it to a shared function?

tpluscode (changes requested):

PR #501, opened 2018-03-16
DNN-18105: don't call _removePageFromTree in duplicate page mode.

zyhfish:


tpluscode (approved):
Looks good to me. Any comments @ricardodiasc ?

PR #502, opened 2018-03-16
Bug/dnn 8327

mean2me:


tpluscode:
Closing this as it will be reassessed what is the root cause of the problem

PR #503, opened 2018-03-23
DNN-9086: Provide Edit as well as Settings menu on Content > Pages > Advanced > Modules screen.

zyhfish:


src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/Modules/ModuleEdit/ModuleEdit.jsx
```diff
@@ -37,21 +47,55 @@ class ModuleEdit extends Component {
...
+            case "settings":
+                editUrl = props.module.editSettingUrl;
+                break;
+        }
+
+        if(editUrl !== ""){
+            if(editUrl.indexOf('popUp') === -1){
+                this.redirectUrl(editUrl);
+            } else {
+                editUrl += (editUrl.indexOf("?") > -1 ? "&" : "?") + queryString;
```
tpluscode: I've see similar snippet a few times recently. We have to stop manipulating query strings that way. This method has already produced a bug not long ago. 

Please add a util which you will use to append the querystrings. Something like

``` js
import { url } from 'the-new-util';

editUrl = url.appendQuerystring(editUrl, { 
      'HideCancel': true, 
      'HideDelete': true, 
      'NoRedirectOnUpdate': true 
   });
```

Here's a snippet using a small library [url-parse](https://www.npmjs.com/package/url-parse) which can be used in implementation:

```js
var URL = require('url-parse');

var url = new URL(
  'https://github.com/foo/bar?test=some',
  /* if needed, pass true to parse query into JS object */ false);
document.write(`<pre>${JSON.stringify( url, null, 2 )}</pre>`);

url.set('query', { 
  'test': 'some', 
  'more': 'params' 
});
document.write(`<pre>${JSON.stringify( url, null, 2 )}</pre>`);
```

tpluscode (changes requested):

tpluscode:
fixed #551

PR #505, opened 2018-03-28
DNN-18242 Pages: When duplicating a page the title field is not clear…

Mohtshum:
…ed for the newly created page

- make sure to clear the temporary added item with the successfully saved page item


tpluscode:
fixed #553

PR #508, opened 2018-04-12
DNN-17335: call to normal schedul update process when schedule enable status changed.

zyhfish:


src/Modules/Settings/Dnn.PersonaBar.TaskScheduler/Services/TaskSchedulerController.cs
```diff
@@ -362,12 +362,13 @@ public HttpResponseMessage UpdateScheduleItem(ScheduleDto scheduleDto)
...
 
-                if ((updatedItem.ScheduleStartDate != existingItem.ScheduleStartDate) ||
-                    (updatedItem.Enabled) ||
-                    (updatedItem.TimeLapse != existingItem.TimeLapse) ||
-                    (updatedItem.RetryTimeLapse != existingItem.RetryTimeLapse) ||
-                    (updatedItem.RetryTimeLapseMeasurement != existingItem.RetryTimeLapseMeasurement) ||
-                    (updatedItem.TimeLapseMeasurement != existingItem.TimeLapseMeasurement))
+                if (updatedItem.ScheduleStartDate != existingItem.ScheduleStartDate ||
+                    updatedItem.Enabled ||
+                    updatedItem.Enabled != existingItem.Enabled ||
```
tpluscode, line 367: Is it relevant to check whether the change is from disabled to enabled or the other way round?

src/Modules/Settings/Dnn.PersonaBar.TaskScheduler/Services/TaskSchedulerController.cs
```diff
@@ -362,12 +362,13 @@ public HttpResponseMessage UpdateScheduleItem(ScheduleDto scheduleDto)
...
+                    updatedItem.TimeLapse != existingItem.TimeLapse ||
+                    updatedItem.RetryTimeLapse != existingItem.RetryTimeLapse ||
+                    updatedItem.RetryTimeLapseMeasurement != existingItem.RetryTimeLapseMeasurement ||
+                    updatedItem.TimeLapseMeasurement != existingItem.TimeLapseMeasurement)
```
tpluscode, line 371: By the way, my head spins from all those statements. Maybe the condition can be simplified?

src/Modules/Settings/Dnn.PersonaBar.TaskScheduler/Services/TaskSchedulerController.cs
zyhfish: change from disabled to enabled should always do the normal update.

src/Modules/Settings/Dnn.PersonaBar.TaskScheduler/Services/TaskSchedulerController.cs
zyhfish: hmm, looks like need to check whether these properties has been changed one by one.

tpluscode: You say that we may be missing doing the wrong update in some cases?

zyhfish: I can't find out a better way to check all these conditions. :(

tpluscode:
fixes #556

PR #510, opened 2018-04-16
DNN-19410

mean2me:


src/Modules/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/webpack.config.js
```diff
@@ -60,7 +60,10 @@ module.exports = {
         })
     ] : [
             new webpack.DefinePlugin({
-                VERSION: JSON.stringify(packageJson.version)
+                VERSION: JSON.stringify(packageJson.version),
+                "process.env": {
+                    "NODE_ENV": JSON.stringify("development")
```
tpluscode: I don't think we really need this constant. It's enough to only ever check for `"production`"

tpluscode (changes requested):
I did notice this some time ago but I cannot recall why we eventually kept the `reduxImmutableStateInvariant()` 🤔 

Also, I see it appears in pretty much [every single of the projects under this repository](https://github.com/search?q=org%3Adnnsoftware+reduxImmutableStateInvariant&type=Code). Should you change every `configureStore.js`?

mean2me:
@tpluscode About *reduxImmutableStateInvariant* there many advices on the web about it. Using in a production environment overhead our application.

tpluscode (approved):

tpluscode:
fixes #557

PR #511, opened 2018-04-17
DNN-19423

mean2me:


src/Modules/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/editLanguagePanel/resourceEditor/index.jsx
```diff
@@ -87,9 +87,7 @@ class ResourceEditor extends Component {
             inFullMode: false
         });
 
-        setTimeout(() => {
-            window.dnn.stopEscapeFromClosingPB = false;
-        }, 500);
+        window.dnn.stopEscapeFromClosingPB = false;
```
tpluscode, line 90: Every other `setTimeout` was wrapping access to a `this`. Here is the only case where the callback accesses the global scope. Isn't this one actually safe?

src/Modules/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/languageSettings/languageVerifier/languageVerifierGrid/index.jsx
```diff
@@ -14,11 +14,9 @@ class LanguageVerifierGridPanel extends Component {
     }
 
     uncollapse(id) {
-        setTimeout(() => {
-            this.setState({
-                openId: id
-            });
-        }, this.timeout);
```
tpluscode, line 21: If you remove `setTimeout` I'd assume that `this.timeout` would also be unnecessary. Where does it come from? I only ever see it being read.

tpluscode (commented):
I searched quickly and didn't immediately find a resource discouraging `setTimeout`within React components. Could you post some?

Still, this code was written this way for a reason. When you remove these delays, won't it change the behaviour of the application?

src/Modules/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/languageSettings/languageVerifier/languageVerifierGrid/index.jsx
mean2me: Looks like it was undefined.

src/Modules/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/editLanguagePanel/resourceEditor/index.jsx
mean2me: it is useless to wrap a variable assignment like this into a setTimeout.

src/Modules/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/languageSettings/languageVerifier/languageVerifierGrid/index.jsx
mean2me: IMO, undefined

mean2me:
After investigating and testing this application I found that setTimeouts are not used for a specific design purpose. They are like manually adjusted in a "slow enough to work" way. 
All starts from the asynchronous way the extension module is loaded. After I fixed all, I couldn't see any regression.

tpluscode:
fixes #559

PR #513, opened 2018-04-18
DNN-19501: get-page command fetching page information which does not …

Mohtshum:
…belong the portalId parament pass for api call


tpluscode (approved):

tpluscode:
fixed #573

PR #514, opened 2018-04-20
Bug/dnn 19500

Mohtshum:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -206,24 +206,32 @@ public TabInfo MovePage(PageMoveRequest request)
...
+            var tab = TabController.Instance.GetTab(page.Id, currentPortal.PortalId);
             if (tab == null)
             {
                 throw new PageNotFoundException();
             }
 
             if (TabPermissionController.CanDeletePage(tab))
             {
-                if (TabController.IsSpecialTab(tab.TabID, portalSettings.PortalId))
+                if (TabController.IsSpecialTab(tab.TabID, currentPortal.PortalId))
```
tpluscode, line 220: Are you sure this check should be done only for the current portal and not the actual portal coming from the request? Or both?

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -206,24 +206,32 @@ public TabInfo MovePage(PageMoveRequest request)
             return TabController.Instance.GetTab(request.PageId, portalSettings.PortalId);
         }
 
-        public void DeletePage(PageItem page)
+        public void DeletePage(PageItem page, PortalSettings portalSettings = null)
         {
-            var portalSettings = PortalController.Instance.GetCurrentPortalSettings();
-            var tab = TabController.Instance.GetTab(page.Id, portalSettings.PortalId);
+            PortalSettings = portalSettings;
```
tpluscode: Is there no side-effect of setting this property?

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -206,24 +206,32 @@ public TabInfo MovePage(PageMoveRequest request)
...
+                if (TabController.IsSpecialTab(tab.TabID, currentPortal.PortalId))
                 {
                     throw new PageException(Localization.GetString("CannotDeleteSpecialPage"));
                 }
                 else
                 {
-                    TabController.Instance.SoftDeleteTab(tab.TabID, portalSettings);
+                    if (PortalHelper.IsContentExistsForRequestedPortal(tab.PortalID, PortalSettings))
+                    {
+                        TabController.Instance.SoftDeleteTab(tab.TabID, currentPortal);
```
tpluscode: The if uses `PortalSettings` and the delete uses `currentPortal`. Is that correct?

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
Mohtshum: Nope this is setting for only DeletePage request path

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
Mohtshum: Actually currentPortal (more intitutive name) is same as protalSettings but it is rename to it as it is getting initialized against same method i.e. PortalController.Instance.GetCurrentPortalSettings();

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
Mohtshum: we are only making sure that if portalId is passed as cross-scripting parameter by 

 if (PortalHelper.IsContentExistsForRequestedPortal(tab.PortalID, PortalSettings))

Meanwhile Delete is getting called with same (renamed) parameter.

tpluscode: I see it's the same thing in all those comments. It's too confusing that there is the property, parameter and local variable.

I understand that all of them should be the same? Please use only the method's params and remove usage of the others from this method.

tpluscode (changes requested):

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
Mohtshum: Now have remove the calls for property PortalSettings object and using the parameter object.

tpluscode (approved):

tpluscode:
fixed #574

PR #515, opened 2018-04-22
DNN-7976 - Make changes on page editing reflect on other components in page modules

ricardodiasc:
This will make changes in visibility be reflected in the tree view and page details.
- Connected components that watch selectedPage with redux state
- Change the PersonaBarTreeview to watch redux state and update visibility based on selectedPage after save
- Minor code improvements


src/Modules/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageDetails/PageUrl/PageExternalUrl/PageExternalUrl.jsx
```diff
@@ -1,4 +1,6 @@
-import React, {Component, PropTypes} from "react";
+import React, {Component} from "react";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
```
tpluscode, line 3: This is something I assume we should avoid. Is `PageExternalUrl` itself a stateless component? Isn't it an issue that we're making it directly dependent on redux?

tpluscode (commented):
Please see my comment. Also, you added tests for `Breadcrumb`. Whys this one in particular? Do we want more, for other modified elements?

ricardodiasc:
The breadcrumb came from another branch. It is already merged, so it does not relate to this commit.

tpluscode:
fixes #560

PR #518, opened 2018-04-25
DNN-19712 : "restore-page" is working regardless of the CrossPortalId…

Mohtshum:
… specified in rest request

- check for Id & name paramter


tpluscode (approved):

tpluscode:
fixes #569

PR #519, opened 2018-04-26
DNN-19566: Set-page working regardless of the CrossPortalId specified…

Mohtshum:
… as parameter

- Add optional parameter of PortalSettings, so that existing functionality should not break
- Make sure to user request PortalSetting if portalId is passed in REST request


tpluscode:
fixes #570

PR #521, opened 2018-04-30
DNN-20005 : Purge command working irrespective of passed crossportal …

Mohtshum:
…id  for rest request


tpluscode:
partially fixes #571

PR #528, opened 2018-05-11
Bug/dnn 19865

Mohtshum:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -556,10 +556,11 @@ private bool ValidatePageUrlSettings(PageSettings pageSettings, TabInfo tab, ref
             {
                 return true;
             }
-
+                        
             bool modified;
             //Clean Url
             var options = UrlRewriterUtils.ExtendOptionsForCustomURLs(UrlRewriterUtils.GetOptionsFromSettings(new FriendlyUrlSettings(portalSettings.PortalId)));
+            urlPath = GetLocalPath(urlPath);
             urlPath = FriendlyUrlController.CleanNameForUrl(urlPath, options, out modified);
```
tpluscode: What I meant for the tests was to check that this method is called with the local path.

src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -1273,6 +1274,16 @@ public void CopyContentFromSourceTab(TabInfo tab, int sourceTabId, IEnumerable<M
             CopyModulesFromSourceTab(tab, sourceTab, includedModules);
         }
 
+        public string GetLocalPath(string url)
```
tpluscode: And this should become private

src/Modules/Content/Dnn.PersonaBar.Pages/Components/IPagesController.cs
```diff
@@ -60,7 +60,7 @@ public interface IPagesController
         PageUrlResult CreateCustomUrl(SeoUrl dto);
         PageUrlResult UpdateCustomUrl(SeoUrl dto);
         PageUrlResult DeleteCustomUrl(UrlIdDto dto);
-
         PagePermissions GetPermissionsData(int pageId);
+        string GetLocalPath(string absoluteURL);
```
tpluscode: This is an implementation detail of the controller. Please remove from interface and verify the call to `FriendlyUrlController.CleanNameForUrl` when calling `ValidatePageUrlSettings`

tpluscode (changes requested):

src/Modules/Content/Dnn.PersonaBar.Pages/Components/IPagesController.cs
Mohtshum: As FriendlyUrlController.CleanNameForUrl is static method call, then is should be wrapped up in an interface call with same arguments and also need to move the ValidatePageUrlSettings to interface and add all dependencies to injection

tpluscode: Oh, you are right, `FriendlyUrlController` is not one of it's not one of those singleton service locators. What a pity.

Still, `GetLocalPath` does not belong to the interface. I concur with the idea to wrap `FriendlyUrlController` and mock that. It will be enough to make `ValidatePageUrlSettings` public but without adding it to the interface. Just use a the concrete type in test

Mohtshum:
Changes pushed for fix in this branch. And the changes for UnitTest would be pushed in [DNN-20362](https://jira.devfactory.com/browse/DNN-20362)

tpluscode (approved):

tpluscode:
partially fixes #572

PR #529, opened 2018-05-15
Bug/dnn 20362

Mohtshum:


tpluscode (commented):
Great, I like it in general.

My only question is about the new interface. Why did you name it `IStaticDependecyResolver`?

Mohtshum:
Actually I want that to use for static method calls for different classes as well. Rather then creating a new Interface for each it could be used to wrap their calls.

tpluscode:
Please don't do that. It will potential create a weird "Frankenstein" interface with many unrelated responsibilities. 

Instead we should have dedicated, possibly small interfaces with well-defined functionality.

tpluscode:
I suppose a natural way would be to have a dedicated wrapper for the wrapped types. In this case it's `FriendlyUrlController` and `UrlRewriterUtils`

tpluscode (approved):
Perfect 👌 

Also, thanks for renaming that parameter. I would have missed it 🥇

tpluscode:
partially fixes #572

PR #530, opened 2018-05-16
DNN-7820: set padding right style has higher priority.

zyhfish:


tpluscode:
Oh, maybe there is a better way than `!important`?

zyhfish:
has to use this as evoq overwrite this style.

tpluscode:
What about a more specific selector? 

Any ideas @ricardodiasc @mean2me ?

tpluscode:
> evoq overwrite this style

Or, maybe the fix belongs to Evoq then?

mean2me:
Probably if we define the rule inside the #users-container selector, it will have priority:
(https://github.com/dnnsoftware/Dnn.AdminExperience.Extensions/blob/da4e4e32e8fd61a1336f4ec7f5b40927bf19af07/src/Modules/Manage/Dnn.PersonaBar.Users/Users.Web/src/components/Body/style.less)

PR #535, opened 2018-05-24
DNN-20661: Make get-page testable

Mohtshum:
- Make get-page dependency injectable 
- Add unit test for success and failure cases

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
+++ b/src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
@@ -1,0 +1,172 @@
+﻿using Moq;
+using NUnit.Framework;
+using DotNetNuke.Entities.Tabs;
+using DotNetNuke.Entities.Portals;
+using Dnn.PersonaBar.Library.Helper;
+using Dnn.PersonaBar.Library.Prompt;
+using Dnn.PersonaBar.Library.Prompt.Models;
+using Dnn.PersonaBar.Pages.Components.Security;
+using Dnn.PersonaBar.Pages.Components.Prompt.Commands;
+
+namespace Dnn.PersonaBar.Pages.Tests
+{
+    [TestFixture]
+    public class GetPageUnitTests
+    {
+        private Mock<ITabController> _tabControllerMock;
+        private Mock<IContentVerifier> _contentVerifierMock;
+        private Mock<ISecurityService> _securityServiceMock;
+
+        [SetUp]
+        public void RunBeforeAnyTest()
+        {
+            _tabControllerMock = new Mock<ITabController>();
+            _securityServiceMock = new Mock<ISecurityService>();
+            _contentVerifierMock = new Mock<IContentVerifier>();
+        }
+
+        [Test]
+        public void Run_GetPageWithValidCommand_ShouldSuccessResponse()
+        {
+            // Arrange
+            var tabId = 21;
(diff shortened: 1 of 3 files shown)


src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            // Assert
+            Assert.IsFalse(result.IsError);
+            Assert.IsNotNull(result.Data);
+            Assert.AreEqual(1, result.Records);
+            Assert.IsFalse(result is ConsoleErrorResultModel);
+
+            _tabControllerMock.Verify(t => t.GetTab(tabId, testPortalId));
+            _securityServiceMock.Verify(s => s.CanManagePage(tabId));
+            _contentVerifierMock.Verify(c => c.IsContentExistsForRequestedPortal(testPortalId, portalSettings, false));
```
tpluscode: I don't think we need mock verification if they are not subject of the test

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            // Arrange
+            var tabId = 212;
+            int testPortalId = 1;
+
+            TabInfo nullTab = null;
+
+            TabInfo tab = new TabInfo();
+            tab.TabID = tabId;
+            tab.PortalID = testPortalId;
```
tpluscode: `tab` is unused in this test.

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            PortalSettings portalSettings = new PortalSettings();
+            portalSettings.PortalId = testPortalId;
+
+            _tabControllerMock.SetReturnsDefault(nullTab);
+            _securityServiceMock.SetReturnsDefault(true);
+            _contentVerifierMock.SetReturnsDefault(false);
```
tpluscode: Common default behaviour can be placed in the `[SetUp]` method to declutter individual tests.

tpluscode (changes requested):

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
Mohtshum: Not exactly same for each test

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
Mohtshum: Mock verification helping the logical flow for test, as if tab null in that case bypass security and content verification. I think it should stay any way I can remove them if you want :(

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/GetPageUnitTests.cs
Mohtshum: While I think you mean setting up of initializing tab, portalSettings and getComands

tpluscode (approved):

PR #536, opened 2018-05-25
DNN-20662: Make restore-page testable

Mohtshum:
- Make changes in restore-page by constructor dependency injection so that behavior of dependencies could be testable
- Add Unit test with success/failure response.

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Pages.Tests/RestorePageUnitTests.cs
+++ b/src/Modules/Content/Dnn.PersonaBar.Pages.Tests/RestorePageUnitTests.cs
@@ -21,7 +21,7 @@ namespace Dnn.PersonaBar.Pages.Tests
         Mock<ITabController> _tabControllerMock;
         Mock<IRecyclebinController> _recyclebinControllerMock;
         Mock<IContentVerifier> _contentVerifierMock;
-        
+
         int _tabId = 91;
         int _testPortalId = 1;
 
@@ -40,16 +40,16 @@ namespace Dnn.PersonaBar.Pages.Tests
             _tabControllerMock = new Mock<ITabController>();
             _recyclebinControllerMock = new Mock<IRecyclebinController>();
             _contentVerifierMock = new Mock<IContentVerifier>();
+
+            _tabControllerMock.SetReturnsDefault(_tab);
+            _contentVerifierMock.SetReturnsDefault(true);
+            _recyclebinControllerMock.Setup(r => r.RestoreTab(_tab, out _message));
         }
 
         [Test]
         public void Run_RestorePage_WithValidCommand_ShouldReturnSuccessResponse()
         {
-            // Arrange            
-            _tabControllerMock.SetReturnsDefault(_tab);
-            _contentVerifierMock.SetReturnsDefault(true);
-            _recyclebinControllerMock.Setup(r => r.RestoreTab(_tab, out _message));
-
+            // Arrange                      
             SetupCommand();
 
             // Act
@@ -59,7 +59,7 @@ namespace Dnn.PersonaBar.Pages.Tests
             Assert.IsFalse(result.IsError);
             Assert.AreEqual(1, result.Records);
             Assert.IsFalse(result is ConsoleErrorResultModel);
-        }    
+        }
 
         [Test]
         public void Run_RestorePage_WithValidCommandForNonExistingTab_ShouldReturnErrorResponse()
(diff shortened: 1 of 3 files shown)


src/Modules/Content/Dnn.PersonaBar.Recyclebin/Components/Prompt/Commands/RestorePage.cs
```diff
@@ -83,8 +87,18 @@ public override ConsoleResultModel Run()
...
+            return string.IsNullOrEmpty(message)
+                ? new ConsoleResultModel(
+                    string.Format(
+                        LocalizeString("Prompt_PageRestoredSuccessfully"),
+                        tab.TabID,
+                        tab.TabName
+                        )
+                    )
+                { Records = 1 }
+                : new ConsoleErrorResultModel(message);
```
tpluscode: This isn't readable now or before. Please change into an `if..else` statement.

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/RestorePageUnitTests.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        [Test]
+        public void Run_RestorePage_WithValidCommand_ShouldReturnSuccessResponse()
+        {
+            // Arrange            
+            _tabControllerMock.SetReturnsDefault(_tab);
+            _contentVerifierMock.SetReturnsDefault(true);
+            _recyclebinControllerMock.Setup(r => r.RestoreTab(_tab, out _message));
+
+            SetupCommand();
```
tpluscode: Why not run `SetupCommand` within `RunBeforeAnyTest`?

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/RestorePageUnitTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            Assert.IsTrue(result.IsError);
+            Assert.IsTrue(result is ConsoleErrorResultModel);
+        }
+
+        [Test]
+        public void Run_RestorePage_WithValidCommandForRequestedPortalNotAllowed_ShouldReturnErrorResponse()
+        {
+            // Arrange
+            _tabControllerMock.SetReturnsDefault(_tab);
+            _contentVerifierMock.SetReturnsDefault(false);
```
tpluscode: I think that negative test should set up the specific method. Otherwise there is a chance that the test will produce inaccurate result (false positive or false negative)

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/RestorePageUnitTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        [Test]
+        public void Run_RestorePage_WithValidCommandForNotRestoreTab_ShouldReturnErrorResponse()
+        {
+            // Arrange
+            _message = "Tab not found";
+
+            _tabControllerMock.SetReturnsDefault(_tab);
+            _contentVerifierMock.SetReturnsDefault(true);
```
tpluscode: Similarly to the other PR, the most common, positive scenario could be set up in the `[SetUp]` method and individual tests can then only override the mock to cater for specific requirement.

```c#
[SetUp]
public void RunBeforeAnyTest()
{
    _tabControllerMock.SetReturnsDefault(_tab);
    _contentVerifierMock.SetReturnsDefault(true);
}

[Test]
public void SomeTest_WhenVerifierSomethingIsFalse()
{
    // local setup, specific to the tested scenario
    _contentVerifierMock.Setup(verifier => verifier.Something()).Return(false);
}
```

This will produce tests with cleaner semantics

tpluscode (changes requested):

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/RestorePageUnitTests.cs
Mohtshum: 💯

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/RestorePageUnitTests.cs
Mohtshum: Its failing for some cases so undo that

Mohtshum:
Changes pushed, same I would do for #535 , if approved or merged

tpluscode (approved):

PR #538, opened 2018-05-29
DNN-20086: GetPageDetails getting crashed with null object reference

Mohtshum:


src/Modules/Content/Dnn.PersonaBar.Pages/Components/PagesControllerImpl.cs
```diff
@@ -1027,14 +1027,15 @@ public IEnumerable<Url> GetPageUrls(int tabId)
 
         public PageSettings GetPageSettings(int pageId, PortalSettings requestPortalSettings = null)
         {
-            var tab = GetPageDetails(pageId);            
+            var tab = GetPageDetails(pageId);
 
-            if (!_contentVerifier.IsContentExistsForRequestedPortal(tab.PortalID, requestPortalSettings))
+            var portalSettings = requestPortalSettings ?? PortalController.Instance.GetCurrentPortalSettings();
```
tpluscode: I think that with this class we've already moved to constructor-injecting dependencies

tpluscode (changes requested):
Also, please add a simple unit test to verify that `PortalController#GetCurrentPortalSettings` is called in case of null parameter

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/PagesControllerUnitTests.cs
```diff
@@ -54,32 +59,60 @@ public void ValidatePageUrlSettings_CleanNameForUrl_URLArgumentShouldBeLocalPath
...
+
+            // Arrange
+            _tabControllerMock.Setup(t => t.GetTab(It.IsAny<int>(), It.IsAny<int>())).Returns(_tab);
+            _portalControllerMock.Setup(p => p.GetCurrentPortalSettings()).Returns(_portalSettings);
+            _contentVerifierMock.Setup(c => c.IsContentExistsForRequestedPortal(It.IsAny<int>(), It.IsAny<PortalSettings>(), false)).Returns(false);
+
+            InitializePageController();
+
+            // Act
+            TestDelegate pageSettingsCall = () =>  _pagesController.GetPageSettings(tabId, null);
```
Mohtshum, line 97: GetPageSettings has lots of dependencies, to minimize the change set to validate the null parameter changes are made for exception throwing case.

tpluscode: Yeah, I see how this avoids all the setup which would be necessary to test a "happy path". It does verify what we had to verify so that's fine

src/Modules/Content/Dnn.PersonaBar.Pages.Tests/PagesControllerUnitTests.cs
```diff
@@ -23,13 +24,16 @@ public class PagesControllerUnitTests
         Mock<IUrlRewriterUtilsWrapper> _urlRewriterUtilsWrapperMock;
         Mock<IFriendlyUrlWrapper> _friendlyUrlWrapperMock;
         Mock<IContentVerifier> _contentVerifierMock;
-
         PagesControllerImpl _pagesController;
+        Mock<IPortalController> _portalControllerMock;
+
+        TabInfo _tab = new TabInfo();
+        PortalSettings _portalSettings = new PortalSettings();
```
tpluscode: I think I's prefer to keep `TabInfo` and `PortalSettings in individual test methods

tpluscode (changes requested):

tpluscode (approved):

PR #539, opened 2018-05-31
Move CI configuration from Jenkins

tpluscode:
This changes AppVeyor configuration so that builds will be triggered for every PR

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,7 @@ before_build:
   - nuget restore src/Dnn.PersonaBar.Extensions.sln -Source https://api.nuget.org/v3/index.json -Source https://www.myget.org/F/dnn-software-public/api/v3/index.json
   - ps: gitversion /l console /output buildserver
 
-build:
+build_script:
   - msbuild src/Dnn.PersonaBar.Extensions.sln /t:Rebuild /v:m /nologo /m /p:Configuration=Release
   - msbuild src/Build/BuildScripts/CreateSourcePackage.build /t:CreateSource /v:m /nologo
   - msbuild src/Build/BuildScripts/CreateSourcePackage.build /t:CreateSymbols /v:m /nologo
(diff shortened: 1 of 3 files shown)


tpluscode:
Please take a look @ahoefling @mitchelsellers

SkyeHoefling:
looks good to me

mitchelsellers:
@ahoefling For tracking, can you do your comment as a true review?

Edit: looks like we don't have proper permissions in this repo...

tpluscode:
Guys, I moved the various scripts that Jenkins currently runs to the AppVeyor's config:

1. Added the yarn script which installs JS packages
1. create nupkg and install packages (added as artifacts)
1. tests now run

One thing you will notice is GitVersion. Don't mind it's now showing `1.5.1`. This would require fixing the branching model first.

tpluscode:
@mitchelsellers please check the permissions again The advisory group team is added with write level.

SkyeHoefling:
The changes look good to me but I am not 100% certain on all the build scripts for this project. I spoke with @mitchelsellers and @ohine and we believe @ohine can handle the review of this one to make sure it is all good.

tpluscode:
> I am not 100% certain on all the build scripts for this project

@ahoefling Are you referring to the msbuild files? Those are the very same scripts which have been used on Jenkins and TeamCity before that. Granted, they could be rewritten to be much simpler but it's nothing new.

ohine (approved):
This looks good to me, but we might be changing how things are packaged as we continue on the CI tasks for the platform.

tpluscode:
Would you merge this then @ohine ?

> we might be changing how things are packaged as we continue on the CI tasks for the platform.

For all I care, we could (temporarily) keep PB builds on AppVeyor and only build Platform on VSTS. We only have to figure out a way to deliver artifacts which works for us. 

That said, migrating from appveyor.yml to other CI should be a piece of cake :D

tpluscode:
I did a mess of commits here by the way. Definitely should squash.

mitchelsellers:
Closing this due to change to VSTS

PR #540, opened 2018-06-03
DNN-20454: Prompt: Cross-portal scripting: "set-user" is not working …

Mohtshum:
…with anything other than the main site (To be general, the site in the URL))

    - Changes made in ValideUser to get the correct or related portalId of the user if accessible
    - Changes made in UnitTest for GetUser for validate user
    - Make Set-User dependencies injectable by constructor 
    - Add unit test for set-user test-cases

### DIFF
--- a/src/Modules/Manage/Dnn.PersonaBar.Users.Tests/SetUserUnitTests.cs
+++ b/src/Modules/Manage/Dnn.PersonaBar.Users.Tests/SetUserUnitTests.cs
@@ -1,5 +1,6 @@
 ﻿using Moq;
 using NUnit.Framework;
+using System.Collections.Generic;
 using DotNetNuke.Entities.Users;
 using DotNetNuke.Entities.Portals;
 using Dnn.PersonaBar.Users.Components;
@@ -15,7 +16,6 @@ namespace Dnn.PersonaBar.Users.Tests
         private Mock<IUsersController> _usersControllerMock;
         private Mock<IUserControllerWrapper> _userControllerWrapperMock;
         private PortalSettings _portalSettings;
-        private SetUser _command;
         private int _testPortalId = 0;
 
         [SetUp]
@@ -32,7 +32,7 @@ namespace Dnn.PersonaBar.Users.Tests
         [Test]
         public void Run_UserIdNull_ReturnErrorResponse()
         {
-            // Arrange
+            // Arrange          
             UserInfo userinfo;
             ConsoleErrorResultModel errorResponse = new ConsoleErrorResultModel();
 
@@ -40,10 +40,11 @@ namespace Dnn.PersonaBar.Users.Tests
                 .Setup(u => u.ValidateUser(-1, _portalSettings, null, out userinfo))
                 .Returns(errorResponse);
 
-            var args = new[] { "set-user", "", "--username", "testusername", "--firstname", "testfirstname", "--lastname", "testlastname" };
-            SetupCommand(args);
+            var args = new[] { "--username", "testusername", "--firstname", "testfirstname", "--lastname", "testlastname" };
+            var command = SetupCommand(string.Empty, args);
+
             // Act
-            var result = _command.Run();
+            var result = command.Run();
 
             // Assert
             Assert.IsTrue(result.IsError);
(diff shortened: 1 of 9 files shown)


src/Modules/Manage/Dnn.PersonaBar.Users.Tests/SetUserUnitTests.cs
```diff
@@ -0,0 +1,99 @@
...
+            // Assert
+            Assert.IsFalse(result.IsError);
+            Assert.AreEqual(1, result.Records);
+        }
+
+        private void SetupCommand(string[] args)
+        {
+            _command = new SetUser(_userValidatorMock.Object, _usersControllerMock.Object, _userControllerWrapperMock.Object);
+            _command.Initialize(args, _portalSettings, null, -1);
+        }
```
tpluscode: Small change, but maybe it will be a little bit cleaner to change this to `private SetUser CreateCommand()` and remove the field.

src/Modules/Manage/Dnn.PersonaBar.Users.Tests/SetUserUnitTests.cs
```diff
@@ -0,0 +1,99 @@
...
+            _userValidatorMock.Setup(u => u.GetValidPortalId()).Returns(_testPortalId);
+            _userValidatorMock
+                .Setup(u => u.ValidateUser(userId, _portalSettings, null, out userInfo))
+                .Returns(errorResponse);
+            _userControllerWrapperMock
+                .Setup(w => w.GetUserById(_testPortalId, userId))
+                .Returns(userInfo);
+
+            var args = new[] { "set-user", userId.ToString(), "--firstname", "user4", "--lastname", "user4", attributeName, attributeValue };
+            SetupCommand(args);
```
tpluscode: I'd consider simplifying the setup method by putting `"set-user"` inside (it's always the same value anyway) and make the variable portion a `params` argument.

```c#
private void SetupCommand(string userId, params string[] args)
{
}
```

And usage would be more readable

```c#
SetupCommand(
    userId.ToString(),
    "--firstname", "user4", 
    "--lastname", "user4", 
    attributeName, attributeValue)
```

src/Modules/Manage/Dnn.PersonaBar.Users/Components/IUserControllerWrapper.cs
```diff
@@ -8,7 +8,7 @@ namespace Dnn.PersonaBar.Users.Components
     public interface IUserControllerWrapper
     {
         UserInfo GetUser(int value, PortalSettings portalSettings, UserInfo currentUserInfo, out KeyValuePair<HttpStatusCode, string> response);
-        UserInfo GetUserById(int portalID, int value);
+        UserInfo GetUserById(int portalId, int userId);
```
tpluscode, line 11: :+1:

tpluscode (changes requested):
I don't like where `UserValidator` is going. The `ValidateUser` method just keeps getting more responsibilities. It now returns the result model, a user through the `out` parameter and modifies its state.

I think we should take a step back here. Is it really a validator? Looks like this entire function should be split

1. Check if `userId.HasValue` belongs to the command itself. Is there a way to do argument validation?
1. `ConsoleErrorResultModel` also belongs to command code and not users/portals code.
1. The method in general is more of `GetUser` than `ValidateUser`. Maybe it should be moved to `UsersController(Wrapper)` and return a DTO combining the `UserInfo` and `PortalId` instead of using `out` param.

src/Modules/Manage/Dnn.PersonaBar.Users.Tests/UserValidatorUnitTests.cs
```diff
@@ -134,7 +134,7 @@ private void SetupForSiteGroup(bool isAllowed, int userId)
             var portals = new ArrayList();
             portals.Add(new PortalInfo() { PortalID = otherPortalId });
             _portalControllerMock.Setup(p => p.GetPortals()).Returns(portals);
-         
+
             _userControllerWrapperMock.Setup(u => u.GetUserById(It.IsAny<int>(), userId)).Returns(userInfo);
             _contentVerifierMock.Setup(c => c.IsContentExistsForRequestedPortal(It.IsAny<int>(), It.IsAny<PortalSettings>(), true)).Returns(isAllowed);
         }
```
Mohtshum, line 140: This file was undone

tpluscode (approved):

tpluscode:
fixes #540

PR #541, opened 2018-06-03
DNN-8316: Prompt Set-user: Firstname and LastName shouldn't be mandatory

Mohtshum:
- Remove the check for true for required for GetFlagValue Call


tpluscode (approved):

tpluscode:
fixed #550

PR #542, opened 2018-06-05
DNN-20873:  delete-user not working across portals and for cross-port…

### LINKED ISSUE #544: delete-user not working across portals and for cross-portals api calls working regardless of API calls
If you pass portal id in rest request to delete user and if user does not belong to that portal, even then it is getting delete by the system

Mohtshum:
…als api calls working regardless of API calls

- API error seems to be fixed by Get-User work
- Mean while User fetched against wrong portalId was giving exception use correct portal Id of user 

fixes #544

### DIFF
--- a/src/Modules/Manage/Dnn.PersonaBar.Users/Components/Prompt/Commands/DeleteUser.cs
+++ b/src/Modules/Manage/Dnn.PersonaBar.Users/Components/Prompt/Commands/DeleteUser.cs
@@ -66,11 +66,11 @@ namespace Dnn.PersonaBar.Users.Components.Prompt.Commands
             // We must clear User cache or else, when the user is 'removed' (so it can't be restored), you 
             // will not be able to create a new user with the same username -- even though no user with that username
             // exists.
-            DotNetNuke.Common.Utilities.DataCache.ClearUserCache(PortalId, userInfo.Username);
+            DotNetNuke.Common.Utilities.DataCache.ClearUserCache(userInfo.PortalID, userInfo.Username);
             // attempt to retrieve the user from the dB
-            userInfo = UserController.GetUserById(PortalId, userInfo.UserID);
+            userInfo = UserController.GetUserById(userInfo.PortalID, userInfo.UserID);
             userModels = new List<UserModel> { new UserModel(userInfo) };
-            return new ConsoleResultModel(LocalizeString("UserDeleted")) { Data = userModels, Records = userModels.Count };
+            return new ConsoleResultModel(LocalizeString("UserDeleted")) { Data = userModels, Records = userModels.Count };            
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/Modules/Manage/Dnn.PersonaBar.Users/Components/Prompt/Commands/DeleteUser.cs
```diff
@@ -56,21 +58,18 @@ public override ConsoleResultModel Run()
                 return new ConsoleErrorResultModel(LocalizeString("Prompt_UserAlreadyDeleted"));
             }
 
-            if (!UserController.DeleteUser(ref userInfo, Notify, false))
+            if (!_userControllerWrapper.DeleteUserAndClearCache(ref userInfo, Notify, false))
```
Mohtshum: How about to save portalId in local variable before passing as ref type parameter?

tpluscode: It would not change anything.

tpluscode: Oh wait, you mean to keep portalId variable. You could do that if you think that the value on the userInfo would change

src/Modules/Manage/Dnn.PersonaBar.Users/Components/Prompt/Commands/DeleteUser.cs
```diff
@@ -56,21 +58,18 @@ public override ConsoleResultModel Run()
...
             }
-            // We must clear User cache or else, when the user is 'removed' (so it can't be restored), you 
-            // will not be able to create a new user with the same username -- even though no user with that username
-            // exists.
-            DotNetNuke.Common.Utilities.DataCache.ClearUserCache(PortalId, userInfo.Username);
-            // attempt to retrieve the user from the dB
-            userInfo = UserController.GetUserById(PortalId, userInfo.UserID);
+
+            // attempt to retrieve the user from the dB 
+            userInfo = _userControllerWrapper.GetUserById(userInfo.PortalID, userInfo.UserID);
```
tpluscode: That said, isn't the purpose of `ref UserInfo` parameter that it shouldn't be necessary to read again? I'd expect it to be returned in its modified state already.

Mohtshum: So, I'm going to put portalId in local variable and then used in above call

Mohtshum: Just found out this method call is changing portalId with MasterId of SiteGroup. So local variable soln. would be save

tpluscode (approved):
LGTM @dnnsoftware/tag

tpluscode:
fixes #567

PR #545, opened 2018-06-10
Bug/dnn 20455

### LINKED ISSUE #548: Prompt: Cross-portal scripting: "restore-user" is not working with anything other than the main site (To be general, the site in the URL))
## Description

Cross-scripting only should work for user if cross-portalId either belongs to user or portalId is one of SiteGroup.
 
 ## Steps to reproduce

- Login as host. 
- From PB, Go Manage, sites. 
- Add new child site (site2) (Will get ID=1) with type Domain (yourMainURL/site2) 
- Login to the child site, add new user; usery (In my case it has ID=5). 
- Delete User 
- Use fiddler 
- Pass the url of you main site +"/API/PersonaBar/Command/Cmd/0" as POST (This targets scripting over the site with id=0) 
- Add all required headers (From the main site's instance) 
- Pass this in the request's body: {"cmdLine=restore-user 5"} 
- Execute the request.

## Current behavior

Successfully restore user

## Expected behavior

User not found

## Affected version

* [x] Platform 9.2

Mohtshum:
fixes #548


tpluscode (commented):
I see you referenced `Dnn.PersonaBar.Users` to get `IUserValidator` here. I suspect this may not be desirable. 

Maybe in such case we should put `IUserValidator` in a shared location? Looks like it only uses Platform library so maybe Dnn.PersonaBar.Prompt would be the right place? @donker 

Unless it's irrelevant in light of dnnsoftware/Dnn.Platform#2090

Mohtshum:
Dnn.PersonaBar.Prompt is the root of Triggering all commands for each User, Page etc. And every of those commands are invoked via reflection from it, It seems moving dependency could create cyclic-dependency issue . Please correct me if missing out the main point? @tpluscode , @donker

sleupold:
the user interfaces and stored procedures should be moved to DNN Library > Entities > Users to be re-usable. Prompt backend and API is also discussed to be moved into DNN Library

Mohtshum:
So we should wait, for this migration of PROMPT to PLATFORM.LIBRARY

Mohtshum:
Having said that, I think currently we should resolve the issue with moving dependency to a new Project Dnn.PersonaBar.Common, which also would help to resolve other PROMPT issues

tpluscode:
@donker I believe we're talking about a bit more distant future now. Other than the reference between projects, which actually we have already between `Roles` and `Users`, do you think we could merge this? More substantial changes would be done when we're reorganising Prompt and we would address this also then? What do you say?

donker:
Why is a user command in the recycle bin when all users logic is in the users module? I suggest moving the command to that module. What do you think, @tpluscode ?

Update: My bad. Then we'd have a reverse dependency. It looks like there is no escaping this link.

So, yes @tpluscode . I think we should continue for now like this. I can't see a simple way out of this unless we're reorganizing code which would be much more invasive and higher risk.

tpluscode (approved):

tpluscode:
Thank you @donker. If you would do the honours and squash/merge this?

ohine (approved):
These changes look good! Thanks!

PR #575, opened 2018-06-15
DNN-20842 : Update yarn.lock to refer only public MyGet repository.

ricardodiasc:
Fixes #575

### DIFF
--- a/src/Modules/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/yarn.lock
+++ b/src/Modules/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/yarn.lock
@@ -2,98 +2,120 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


bdukes:
@ricardodiasc I'm seeing a lot of version changes in the lock file, I'm not seeing URL changes.  Can you specify what's supposed to be changing here?

tpluscode:
See for example the resolution for `enzyme-adapter-react-15` (Line 2040) @bdukes. 

The lock file was generated with `npmrc` using the private NPM repository. One mirrors the other hence the resolution URLs. This then cannot be installed on CI because it only accesses the public feeds.

bdukes:
Got it.  👍

bdukes (approved):
I'd prefer just the URL changes, and version differences being a separate commit, but I don't think it's going to cause any issues.

tpluscode (approved):

PR #582, opened 2018-06-26
add yarn msbuild task

ohine:
This is part of larger pull request from platfrom https://github.com/dnnsoftware/Dnn.Platform/pull/2137 and
https://github.com/dnnsoftware/Dnn.AdminExperience.Library/pull/155

### DIFF
--- a/src/Modules/Content/Dnn.PersonaBar.Recyclebin/Module.build
+++ b/src/Modules/Content/Dnn.PersonaBar.Recyclebin/Module.build
@@ -5,6 +5,7 @@
       <DNNFileName>Dnn.PersonaBar.Recyclebin</DNNFileName>
       <PackageName>Dnn.PersonaBar.Recyclebin</PackageName>
       <ModuleFolderName>$(WebsitePath)\DesktopModules\admin\Dnn.PersonaBar\Modules\Dnn.Recyclebin</ModuleFolderName>
+	  <YarnWorkingDirectory></YarnWorkingDirectory>
   </PropertyGroup>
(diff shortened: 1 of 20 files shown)


src/Modules/Content/Dnn.PersonaBar.Recyclebin/Module.build
```diff
@@ -5,6 +5,7 @@
       <DNNFileName>Dnn.PersonaBar.Recyclebin</DNNFileName>
       <PackageName>Dnn.PersonaBar.Recyclebin</PackageName>
       <ModuleFolderName>$(WebsitePath)\DesktopModules\admin\Dnn.PersonaBar\Modules\Dnn.Recyclebin</ModuleFolderName>
+	  <YarnWorkingDirectory></YarnWorkingDirectory>
```
bdukes, line 8: @ohine should this have a value?

bdukes (changes requested):

src/Modules/Content/Dnn.PersonaBar.Recyclebin/Module.build
ohine: Some of the packages do not have the packages.json file which yarn is looking for, I was having a difficult time getting it to check if the file exists with the wildcard search, so my fix was to check the length of the working directory property. If it's empty Yarn doesn't try to run on that project.

bdukes (approved):

mitchelsellers (approved):
LGTM

PR #585, opened 2018-07-03
DNN-20456: Cross-portal scripting:"add-roles" is not working

### LINKED ISSUE #586: Prompt: Cross-portal scripting: "add-roles" is not working with anything other than the main site (To be general, the site in the URL))
## Steps to reproduce

- Login as host. 
- From PB, Go Manage, sites. 
- Add new child site (site2) (Will get ID=1) with type Domain (yourMainURL/site2) 
- Login to the child site, add new user; usery (In my case it has ID=5). 
- Use fiddler 
- Pass the url of you main site +"/API/PersonaBar/Command/Cmd/0" as POST (This targets scripting over the site with id=0) 
- Add all required headers (From the main site's instance) 
- Pass this in the request's body: `{"cmdLine=add-roles 5 --roles editor "}`
- Execute the request.

## Current behavior

Successfully role added to the user

## Expected behavior

User not found

Mohtshum:
- In Add-roles, correct the portalId for request, earlier only the main portalId was passing from which the user id logged-in
- In UserController, also correct the portalsettings object

fixes #586

### DIFF
--- a/src/Modules/Manage/Dnn.PersonaBar.Users.Tests/AddRolesUnitTests.cs
+++ b/src/Modules/Manage/Dnn.PersonaBar.Users.Tests/AddRolesUnitTests.cs
@@ -1,11 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using Moq;
 using NUnit.Framework;
 using DotNetNuke.Entities.Users;
 using DotNetNuke.Security.Roles;
-using Dnn.PersonaBar.Library.Prompt;
 using Dnn.PersonaBar.Roles.Components;
 using Dnn.PersonaBar.Users.Components;
 using Dnn.PersonaBar.Users.Components.Prompt.Commands;
@@ -14,19 +12,19 @@ using Dnn.PersonaBar.Library.Prompt.Models;
 namespace Dnn.PersonaBar.Users.Tests
 {
     [TestFixture]
-    public class AddRolesUnitTests : CommandTests
+    public class AddRolesUnitTests : CommandTests<AddRoles>
     {
         private Mock<IUserValidator> _userValidatorMock;
         private Mock<IUsersController> _usersControllerMock;
         private Mock<IRolesController> _rolesControllerMock;
 
-        [SetUp]
-        public void RunBeforeEveryTest()
+        protected override string CommandName { get { return "Add-roles"; } }
+
+        protected override void ChildSetup()
         {
             _userValidatorMock = new Mock<IUserValidator>();
             _usersControllerMock = new Mock<IUsersController>();
             _rolesControllerMock = new Mock<IRolesController>();
-            Reset();
         }
 
         [TestCase]
@@ -73,7 +71,7 @@ namespace Dnn.PersonaBar.Users.Tests
                 .Returns(rolesList);
 
             // Act
-            var result = SetupCommandForAddRolesAndRun(userId.ToString(), "--roles", "Tester");
+            var result = RunCommand(userId.ToString(), "--roles", "Tester");
 
             // Assert
             Assert.IsFalse(result.IsError);
@@ -93,7 +91,7 @@ namespace Dnn.PersonaBar.Users.Tests
                 .Returns(errorResultModel);
 
             // Act
-            var result = SetupCommandForAddRolesAndRun(userId.ToString());
+            var result = RunCommand(userId.ToString());
 
             // Assert
             Assert.IsTrue(result.IsError);
@@ -143,23 +141,15 @@ namespace Dnn.PersonaBar.Users.Tests
                 .Returns(rolesList);
 
             // Act
-            TestDelegate ex = () =>
-            SetupCommandForAddRolesAndRun(userId.ToString(), "--roles", "Not Tester");
+            TestDelegate ex = () => RunCommand(userId.ToString(), "--roles", "Not Tester");
 
             // Assert
             Assert.Throws<Exception>(ex, "Should throw exception");
         }
 
-        protected override ConsoleCommandBase DependencyInitializer()
+        protected override AddRoles CreateCommand()
         {
             return new AddRoles(_userValidatorMock.Object, _usersControllerMock.Object, _rolesControllerMock.Object);
-        }
-
-        private ConsoleResultModel SetupCommandForAddRolesAndRun(params string[] args)
-        {
-            args = new[] { "add-roles" }.Concat(args).ToArray();
-            command = SetupCommand(args);
-            return command.Run();
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Modules/Manage/Dnn.PersonaBar.Users.Tests/CommandTests.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using Dnn.PersonaBar.Library.Prompt;
+using DotNetNuke.Entities.Portals;
+using Dnn.PersonaBar.Library.Prompt.Models;
+using DotNetNuke.Entities.Users;
+
+namespace Dnn.PersonaBar.Users.Tests
+{
+    public abstract class CommandTests
```
Mohtshum: CommandTest will help to achieve DRY for existing tests as couple of properties and methods are used repeatedly like PortalSettings, ErrorModelResult, GetUserInfo and also good to abstract the "Act" part for these test as more or less it is same except for the arguments and type of ConsoleCommand i.e RestoreUser, GetUser etc

src/Modules/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
```diff
@@ -534,6 +541,15 @@ private static bool IsAdmin(UserInfo user, PortalSettings portalSettings)
...
         }
 
+        private PortalSettings GetPortalSettings(int portalId)
+        {
+            var portalSettings = new PortalSettings(portalId);
+            var portalAliases = PortalAliasController.Instance.GetPortalAliasesByPortalId(portalId);
+            portalSettings.PrimaryAlias = portalAliases.FirstOrDefault(a => a.IsPrimary);
+            portalSettings.PortalAlias = PortalAliasController.Instance.GetPortalAlias(portalSettings.DefaultPortalAlias);
+            return portalSettings;
+        }
```
tpluscode, line 551: Looks like code which we could have in other PB commands. If so, then maybe it should be extracted into a common library?

I'd go for an extension method

```c#
public static PortalSettings GetPortalSettings(the IPortalAliasController controller, int portalId)
{
}
```

src/Modules/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
```diff
@@ -534,6 +541,15 @@ private static bool IsAdmin(UserInfo user, PortalSettings portalSettings)
                     includeSuperUsers);
         }
 
+        private PortalSettings GetPortalSettings(int portalId)
+        {
+            var portalSettings = new PortalSettings(portalId);
+            var portalAliases = PortalAliasController.Instance.GetPortalAliasesByPortalId(portalId);
```
tpluscode, line 547: You are using `PortalAliasController.Instance`. Please turn it into instance field. Isn't is also necessary to put in constructor to mock in tests?

src/Modules/Manage/Dnn.PersonaBar.Users.Tests/CommandTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        protected ConsoleCommandBase command;
+        protected PortalSettings portalSettings;
+        protected ConsoleErrorResultModel errorResultModel;
+
+        protected int testPortalId = 0;
+
+        protected abstract ConsoleCommandBase DependencyInitializer();
```
tpluscode: That is a funky name. Would `CreateCommand` or `GetCommand` work? 
Or even `protected abstract ConsoleCommandBase Commend { get; }`

src/Modules/Manage/Dnn.PersonaBar.Users.Tests/CommandTests.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Dnn.PersonaBar.Users.Tests
+{
+    public abstract class CommandTests
+    {
+        protected ConsoleCommandBase command;
+        protected PortalSettings portalSettings;
+        protected ConsoleErrorResultModel errorResultModel;
+
+        protected int testPortalId = 0;
```
tpluscode: Let's avoid non-`private` fields. Are those field necessarily protected? Are they even required? If yes, then let's make them into properties

src/Modules/Manage/Dnn.PersonaBar.Users.Tests/AddRolesUnitTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        private Mock<IUsersController> _usersControllerMock;
+        private Mock<IRolesController> _rolesControllerMock;
+
+        [SetUp]
+        public void RunBeforeEveryTest()
+        {
+            _userValidatorMock = new Mock<IUserValidator>();
+            _usersControllerMock = new Mock<IUsersController>();
+            _rolesControllerMock = new Mock<IRolesController>();
+            Reset();
```
tpluscode: This isn't ideal, because someone writing tests must be aware of this method. Instead I suggest creating a public setup method in base class and an abstract setup method for the inheriting fixture

```c#
public class CommandTests
{
   [SetUp]
   public void Setup()
   {
       // what Reset() does
       ChildSetup();
   }

   protected abstract void ChildSetup();
}

public class AddRolesUnitTests : CommandTests
{
   protected override void ChildSetup()
   {
      _userValidatorMock = new Mock<IUserValidator>(); // etc
   }
}
```

src/Modules/Manage/Dnn.PersonaBar.Users.Tests/CommandTests.cs
tpluscode: Good idea. May I suggest making it generic

```c#
public abstract class CommandTests<TCommand> where TCommand : ConsoleCommandBase
{
}
```

src/Modules/Manage/Dnn.PersonaBar.Users.Tests/AddRolesUnitTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        protected override ConsoleCommandBase DependencyInitializer()
+        {
+            return new AddRoles(_userValidatorMock.Object, _usersControllerMock.Object, _rolesControllerMock.Object);
+        }
+
+        private ConsoleResultModel SetupCommandForAddRolesAndRun(params string[] args)
+        {
+            args = new[] { "add-roles" }.Concat(args).ToArray();
+            command = SetupCommand(args);
+            return command.Run();
```
tpluscode: I think it's a bit repetitive and unnecessary. I would shift more of this to the base class

```c#
private ConsoleResultModel SetupCommandForAddRolesAndRun(params string[] args)
{
   args = new[] { "add-roles" }.Concat(args).ToArray();
   return base.RunCommand(args);
}
```

where `base.RunCommand()` would replace `base.SetupCommand()`

tpluscode (changes requested):

src/Modules/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
Mohtshum: Currently unit test is added for Add-Roles, I think it should be done in separate task

tpluscode (approved):

mitchelsellers (approved):
LGTM

PR #591, opened 2018-07-17
Extensions - Resource dropdown is not populated

### LINKED ISSUE #592: Extensions - Resource dropdown is not populated
## Steps

- Login to Evoq Engage
- navigate to Settings>Extensions
- Click on Create new module
- Select Manifest from the dropdown
- Add Owner Folder & Moldule Folder
- Click on Resource

## Expected

List is displayed for all *.dnn and *.dnnVERSION_NUMBER files in the module folder.

## Current

Resource dropdown is not populated, no matter what combination of Owner Folder & Module folder are selected. Therefore cannot create Manifest as Resource is required field

screenshot-1-2

mean2me:
fixes #592 

### Steps
- Login to Evoq Engage 
- navigate to Settings>Extensions 
- Click on Create new module 
- Select Manifest from the dropdown 
- Add Owner Folder & Moldule Folder 
- Click on Resource

### Current
Resource dropdown is populated, no matter what combination of Owner Folder & Module folder are selected.  Therefore cannot create Manifest as Resource is required field

### Expected
List is displayed for all *.dnn and *.dnnVERSION_NUMBER files in the module folder.

### DIFF
--- a/src/Modules/Settings/Dnn.PersonaBar.Extensions/Services/ExtensionsController.cs
+++ b/src/Modules/Settings/Dnn.PersonaBar.Extensions/Services/ExtensionsController.cs
@@ -65,6 +65,7 @@ namespace Dnn.PersonaBar.Extensions.Services
     public class ExtensionsController : PersonaBarApiController
     {
         private static readonly ILog Logger = LoggerSource.Instance.GetLogger(typeof(ExtensionsController));
+        private static readonly Regex ManifestExensionsRegex = new Regex(@"dnn\d*$");
         private readonly Components.ExtensionsController _controller = new Components.ExtensionsController();
         private static readonly string[] SpecialModuleFolders = new[] { "mvc" };
         private const string AuthFailureMessage = "Authorization has been denied for this request.";
@@ -983,8 +984,7 @@ namespace Dnn.PersonaBar.Extensions.Services
                         files.AddRange(GetFiles(Globals.HostMapPath + "Templates\\", ".module.template"));
                         break;
                     case FileType.Manifest:
-                        var regex = new Regex(@"\.dnn(\d+)?$");
-                        files.AddRange(GetFiles(folder, "*.*").Where(f => regex.IsMatch(f)));
+                        files.AddRange(GetFiles(folder, "*.dnn*").Where(file => ManifestExensionsRegex.IsMatch(file)));
                         break;
                 }
 
--- a/src/Modules/Settings/Dnn.PersonaBar.Extensions/Services/ExtensionsController.cs
+++ b/src/Modules/Settings/Dnn.PersonaBar.Extensions/Services/ExtensionsController.cs
@@ -29,6 +29,7 @@ using System.Net.Http.Formatting;
 using System.Net.Http.Headers;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Web.Http;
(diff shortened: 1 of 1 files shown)


src/Modules/Settings/Dnn.PersonaBar.Extensions/Services/ExtensionsController.cs
```diff
@@ -982,8 +983,8 @@ public HttpResponseMessage GetModuleFiles(string ownerFolder, string moduleFolde
                         files.AddRange(GetFiles(Globals.HostMapPath + "Templates\\", ".module.template"));
                         break;
                     case FileType.Manifest:
-                        files.AddRange(GetFiles(folder, "*.dnn"));
-                        files.AddRange(GetFiles(folder, "*.dnn5"));
+                        var regex = new Regex(@"\.dnn(\d+)?$");
+                        files.AddRange(GetFiles(folder, "*.*").Where(f => regex.IsMatch(f)));
```
tpluscode: Pattern `*.dnn?` should be just fine here, where question mark [substituses a single, optional character](https://msdn.microsoft.com/en-us/library/wz42302f(v=vs.110).aspx#Remarks)

tpluscode:
@mean2me please complete signing the CLA

src/Modules/Settings/Dnn.PersonaBar.Extensions/Services/ExtensionsController.cs
tpluscode: I later realised that `?` will also allow non-digit character. To be 100% correct we could combine both to avoid iterating unnecessary files (is it really necessary) and the only accepting `dnn\d*`. The reg exp can actually be simplified too

```c#
var regex = new Regex(@"dnn\d*$");
files.AddRange(GetFiles(folder, "*.dnn*").Where(file => regex.IsMatch(file)));
```

Simplest possible regex is to check for optional numeric suffix once only `.dnn` file have been selected initially.

Am I overthinking this @dnnsoftware/tag? :)

src/Modules/Settings/Dnn.PersonaBar.Extensions/Services/ExtensionsController.cs
```diff
@@ -982,8 +983,8 @@ public HttpResponseMessage GetModuleFiles(string ownerFolder, string moduleFolde
                         files.AddRange(GetFiles(Globals.HostMapPath + "Templates\\", ".module.template"));
                         break;
                     case FileType.Manifest:
-                        files.AddRange(GetFiles(folder, "*.dnn"));
-                        files.AddRange(GetFiles(folder, "*.dnn5"));
```
tpluscode, line 986: Also, please move the regex to a static readonly field and avoid single-character variables, even in lambdas.

tpluscode (changes requested):

tpluscode (approved):

tpluscode:
@dnnsoftware/tag please review

mitchelsellers (approved):
LGTM

ohine:
@tpluscode can you please add the correct milestone when merging any pull requests?

tpluscode:
I do remember, usually. The milestone is on the issue and no the PR. Should we have them on both @ohine?

ohine:
@tpluscode I think it makes sense to have them on both. It helps determine who contributed to a release vs reported an issue. Some times the reporter isn't the same as the person who contributed the fix. 

I am open to any suggestions to help make release management easier if this doesn't make sense to have them on both.

PR #597, opened 2018-07-20
DNN-21870: only clean user count cache during create user process.

zyhfish:


ashishpd:
This PR is to resolve a major perf problem where anytime a user gets created, DNN flushes the entire entity cache. This causes slow down for sites that have large objects in the entity cache.

tpluscode:
Reported as dnnsoftware/Dnn.Platform#2178

mitchelsellers (approved):
This is a fantastic change, approved!

ohine (changes requested):
I'm curious why we didn't use the helper function `ClearUserCache` to clear this? I'm not a fan of going directly to the CachingProvider and clearing individual keys.

From the looks of it, we should add this key into the helper function and then make sure we're calling that instead. 

https://github.com/dnnsoftware/Dnn.Platform/blob/6387dd1a7e55cdd994d567cdb76733fbd63819dc/DNN%20Platform/Library/Common/Utilities/DataCache.cs#L497-L501

tpluscode:
@zyhfish please reply to Oliver

mitchelsellers:
@ohine Interesting, it looks like the cache key that is getting cleared in this case is for something that isn't cleared when a user's cache is cleared, thus the reason for the direct call. However I want to confirm with @zyhfish that I'm correct in that assumption

zyhfish:
Hi @ohine , ClearUserCache doesn't work here as the user won't have anything cached as he just register as a new user, we only need to clear the user count cache for this situation.
also we can't add this key into ClearUserCache as this method used to clear user-level caches, but UserCountCache is a portal-level cache.

thanks.

tpluscode:
@ohine does Ben's reply satisfy you?

ohine:
@zyhfish that makes sense. If we find out we need to start clearing any more then this single value, we should create a new wrapper function at that point. 

@tpluscode Switching to approve and merge this PR shortly.

ohine (approved):

PR #598, opened 2018-07-20
Fix: missing reference to CallCustomAction

mean2me:
### Description
Impossible to open Analytics from Pages treeview due to a missing reference to CallCustomAction method defined in parent component App.jsx

### Solution
CallCusotmAction method set as a property of PersonaBarPageTreeviewInteractor


tpluscode:
@dnnsoftware/tag looks like this was branched off development, where IMO this should get fixed in 9.2.1.

Please reopen if you think otherwise. If you agree, let's ask @mean2me to submit this for the `release/1.6.0` branch

mitchelsellers:
This change appears to contain a lot more than the actual fix referenced as well.  I would expect this to be a single item change.

tpluscode:
You're right @mitchelsellers. The actual fix is only the last commit 2bfe347 but the PR was initially to development, when we need this in 1.6.0. Unfortunately switching the PR back did not have the desired effect on GitHub

PR #601, opened 2018-07-22
Fix: missing reference to CallCustomAction

mean2me:
### Description
Impossible to open Analytics from Pages treeview due to a missing reference to CallCustomAction method defined in parent component App.jsx

### Solution
CallCusotmAction method set as a property of PersonaBarPageTreeviewInteractor


tpluscode:
fixes #602

tpluscode (approved):

mitchelsellers (approved):
LGTM

PR #610, opened 2018-08-08
DNN-22335: To update the dnn-file-upload control in site-settings

Mohtshum:
Update dnn-file-upload to 0.1.22 (which does contain changes for sending portalid when browse for file based on selected portal)


tpluscode (approved):

tpluscode:
@dnnsoftware/tag this is to apply the fix from dnnsoftware/Dnn.React.Common#144 which we earlier failed to see through.

Notice how every PB extensions is committed with its respective react bundle. In this case we updated the `site-settings-bundle.js` file. I think it's not possible any more now that we have yarn run during build, correct?

Also, we've take the liberty of publishing version `0.1.22` of file upload component. Let's consider automating this process. Until now it was always a manual step.

ohine:
@Mohtshum can you please resubmit this on the combined repository at https://github.com/dnnsoftware/Dnn.AdminExperience

tpluscode:
resubmitted as dnnsoftware/Dnn.AdminExperience#17

